PR #154, opened 2011-11-02
Video model with the possibility of automatic conversion to other formats and dimensions

nmfm:
This is based on a solution we implemented with django for a client. We integrated the code into django-filer because it seemed useful and we could use it in future projects.

Videos can be used without activating the extra functionalities - they will behave just like the normal File object.

We tried to keep a similar architecture to the other django-filer + thumbnailer models.

There's a new Video model based on file, additional settings for configuring the processing with ffmpeg, a couple of management commands to do the conversion work and cleanup unused files, a template tag that renders an html5 video tag using multiple formats to work in different browsers, including legacy flash support (we expect this could be a typical use). Also included are some tests and documentation for the new features.

We are also working on a plugin (HTML5 Video) for django-cms that will make use of these new functionalities (https://github.com/nmfm/cmsplugin-filer).


FinalAngel:
@nmfm as it seems there is not much progress on this feature. Please re-open if its relevant again to the current develop branch.

PR #361, opened 2013-12-21
[WIP] file level permissions

stefanfoulis:
- adds File-level permissions
- simpler permissions: `can_read` (view, download) and `can_edit` (create subfolders, change, delete).
- no more special `can_add_subfolders` or inheritance settings. Folders now always inherit from ancestors.
- smart permission priorities: "closer" permissions override "farther" ones. deny overrides allow. more specific (e.g user) overrides more general (group)
- Permissions are now saved in text fields and are de-normalized on all Folders and Files (zero db-query to check permissions for a user on a file/folder). Saving (changing permissions) is now much more expensive. But reading is cheaper.
- easier to get querysets for a specific permission (without expensive joins)
### unfinished business
- [ ] fix permission queryset to really behave the same as .check()
- [ ] add docs
- [ ] add/fix tests
- [ ] add migration from old permission system
- [ ] replace Permission model with a custom widget for the permission fields
- [ ] optimise performance of permission propagation to descendants
- [ ] handle case where FILER_PERMISSIONS == False

NICE TO HAVE (maybe separate or after frontend refactor)
- [ ] show inherited permissions on folders
- [ ] explain permissions (why a user/group has access)
- [ ] group permissions smartly (everyone first, then groups, then users)
- [ ] allow defining a certain storage backend (public/private) for a folder

WIP:
- [x] add admin ui to files
- [x] add permission propagation to files
- [x] add queryset filtering with PermissionSet and PermissionUser
- [x] add filtering of list views in admin
- [x] add filtering of search views
- [x] handle permission refresh when moving folders
- [x] handle permission refresh when moving file (to clipboard, to other folder)
- [x] handle "root" permissions
- [x] check filer.admin.tools for permission stuff that needs refactoring
- [x] handle permission delete
- [x] update "save_as" paramater on server


FinalAngel:
@stefanfoulis please reopen and rebase if still relevant

PR #463, opened 2014-11-14
generic JS usable when a row is duplicated when file is a part of a ManyToMany

DylannCordel:
How to reproduce :

Use a FilerField in a ManyToMany, and use admin's inline to manage it in the admin form.
When "add new row" is clicked, the "clear" image is displayed and not functionnal.

This is because the old javascript is for the current ID, so new lines, with != IDs, couldn't work. This bug fix use classes and the jQuery(document).on function to watch event on newly created lines.

### DIFF
--- a/filer/static/filer/js/widget.js
+++ b/filer/static/filer/js/widget.js
@@ -17,4 +17,4 @@
         //if this file is included multiple time, we ensure that filer_clear is attached only once.
         $(document).off('click.filer', '.filerFile .filerClearer', filer_clear).on('click.filer', '.filerFile .filerClearer', filer_clear);
     });
-})(jQuery);
+})(django.jQuery);
--- a/filer/static/filer/js/widget.js
+++ b/filer/static/filer/js/widget.js
@@ -1,0 +1,20 @@
+(function($) {
+    var filer_clear = function(e){
+        var clearer = $(this),
+            hidden_input = clearer.closest('.filerFile').find('input.vForeignKeyRawIdAdminField'),
+            base_id = '#'+hidden_input.attr('id'),
+            thumbnail = $(base_id+'_thumbnail_img'),
+            description = $(base_id+'_description_txt'),
+            static_prefix = clearer.attr('src').replace('admin/img/icon_deletelink.gif', 'filer/');
+        clearer.hide();
+        hidden_input.removeAttr("value");
+        thumbnail.attr("src", static_prefix+"icons/nofile_48x48.png");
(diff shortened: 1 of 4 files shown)


filer/static/filer/js/widget.js
```diff
@@ -0,0 +1,20 @@
...
+        thumbnail.attr("src", static_prefix+"icons/nofile_48x48.png");
+        description.html("");
+    }
+
+    $(document).ready(function(){
+        $('.filerFile .vForeignKeyRawIdAdminField').attr('type', 'hidden');
+        //if this file is included multiple time, we ensure that filer_clear is attached only once.
+        $(document).off('click.filer', '.filerFile .filerClearer', filer_clear).on('click.filer', '.filerFile .filerClearer', filer_clear);
+    });
+})(jQuery);
```
yakky: This should be ``django.jQuery`` to use the django-namespaced jQuery
Could you fix that?

yakky:
Other than the above comment, LGTM, thanks @DylannCordel

DylannCordel:
`django.jQuery` is now used in `filer/static/filer/js/widget.js` and `filer/static/filer/js/popup_handling.js`

yakky:
awesome!

yakky:
Thanks @DylannCordel

DylannCordel:
Hi @evildmp 

I'll add a new paper marked "IE" in my "angry box"... Fixed in #568

Straight from the (recent) jQuery source :

```javascript
// We can't allow the type property to be changed (since it causes problems in IE)
if ( name == "type" && jQuery.nodeName( elem, "input" ) && elem.parentNode )
    throw "type property can't be changed";
```

IE, Glue eater

DylannCordel:
Hi @stefanfoulis 

You right, but this is corrected since 10 january on this commit : 7b5fcc1
See https://github.com/stefanfoulis/django-filer/pull/463#issuecomment-69449705

evildmp:
@DylannCordel The issue is affecting me in Safari though. With the change in #568 I now see:

```
[Error] TypeError: undefined is not a function (evaluating '$(document).off('click.filer', '.filerFile .filerClearer', filer_clear)')
	(anonymous function) (widget.js, line 18)
	ready (jquery.js, line 392)
	DOMContentLoaded (jquery.js, line 745)
```

evildmp:
I should add that the Filer widget itself now appears to work, even though I get that error, but I still find that it breaks other JS that my admin relies on.

PR #474, opened 2014-12-13
Skip new migration if table already exists

yakky:
Skip the new 0002_image django 1.7 migration if table already exists for compatibility with previous migration structure

### DIFF
--- a/filer/migrations_django/0002_image.py
+++ b/filer/migrations_django/0002_image.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from django.db import models, migrations
+from django.db import models, migrations, connection
 from filer.settings import FILER_IMAGE_MODEL
 
 
@@ -13,7 +13,7 @@ class Migration(migrations.Migration):
     ]
 
     operations = []
-    if not FILER_IMAGE_MODEL:
+    if not FILER_IMAGE_MODEL and 'filer_image' not in connection.introspection.table_names():
         operations.append(
             migrations.CreateModel(
                 name='Image',


yakky:
Merging this as-is for release in 0.9.9
More work will go in swappable filer models

jsma:
Yikes. Was there seriously nothing in my previous comment that warranted consideration/comment?

yakky:
It's been duly noted, and actually your comments together with other considerations made me consider a more in depth approach to swappable models.
I neede to merge this to get 0.9.9 out with this stop-gap solution, waiting for a refactoring of swappable models.

jsma:
Care to explain this 'more in depth approach'? Is there some security angle that prevents you from discussing this in an open forum such as the mailing list or on a related Github ticket? Aside from a parochial commercial interest, what made it necessary to "get 0.9.9 out with this stop-gap solution" before a proper solution could be arrived at? I don't recall any demands from the community that this be merged immediately. In fact, I don't recall this being discussed by the community at all.

I've enjoyed our brief interactions online (thanks for merging my many pull requests!) and will happily buy you many beers if we meet in person. Just scratching my head at the moment and pinning my requirements to the version before this was merged. I'm on IRC often. Feel free to yell at me there, here, or anywhere. Cheers.

yakky:
There is no "parochial commercial interest" actually: it's just that 0.9.8 (or any filer version, for that matter) it's not installable with pip 6 / setuptools > 8 due to strict PEP440 implementaion of these tools, and this is breaking user installs.
Merging this it's way better than the solution we have now, so I think it's in the best interest of the project to get this merged and get 0.9.9 out soon (tomorrow, probably) and discuss a better approach to swappable models in other issues / PR.
I'll happily discuss the more in depth approach as soon as I've a clearer picture of it in my mind: but basically it's about making File / Image entirely swappable

jsma:
Apologies for being a jerk in a moment of frustration. Didn't realize the pip 6 issue. Can we revert the abstract image changes in develop and keep them in a feature branch until a correct solution can be found? As-is, yes pip6 users can install the package but the package is broken for anyone that's been using this plugin with Django 1.7 for the last few months and it's also broken for users of cmsplugin-filer due to migration dependencies.

yakky:
@jsma could you point me to the relevant information about such issues?
I tried a 1.7 blank project using filer and cmsplugin_filer with no issues in set it up

jsma:
The Django 1.7 issue is strictly limited to anyone using Django 1.7, like myself, who started using the develop branch a few months ago to get the original Django 1.7-compatible 0001 initial migration. Now the initial migration has changed and the creation of the Image model has been moved to 0002. I still have a strong preference that we alter the 0001 to make image swappable rather than create the 0002, but at a minimum a note in the release docs that anyone like me will have to fake the 0002 would be nice. 

As far as cmsplugin-filer, you'll have to enable the image plugin in your test project. This plugin points at 0001, expecting the Image model to have been created in that migration https://github.com/stefanfoulis/cmsplugin-filer/blob/develop/cmsplugin_filer_image/migrations_django/0001_initial.py

PR #482, opened 2015-01-09
Fix 17 migrations - hopefully

yakky:
This still is not the final solution for the swappable model, but still should be better implementation of the migrations.
Real solution will imply the use of swappable meta attribute, but being still undocumented it may be tricky to properly implement it


yakky:
@jsma care to review this?

yakky:
Ping @jsma would you review if this solution fixes the issues with Django 1.7?

jsma:
Sorry, literally in the middle of launching a big project. Might be able to regroup this weekend.

yakky:
Thanks!

jsma:
Very belated reply but can confirm no issues so far after upgrading to the latest on develop branch.

PR #486, opened 2015-01-17
Remove is_hidden attribute for FilerFolderField

### LINKED ISSUE #485: [Django>=1.7] FilerFolderField admin widget hidden
If you use Django 1.7 and have a FilerFolderField in a model, the corresponding admin widget will not show up.

That’s happening since django/django@dc3d2ac98c1bcfad74d3e9523caf07e7e9fb15aa. In `fieldset.html`, a `hidden` class is added to a form row if all the contained widgets have `is_hidden = True`.

Setting `is_hidden` to `False` works fine. I’m not sure whether this attribute is useful somewhere else. I found it used 7 times in Django, and always refer to showing the field and its label. Since it’s not really a hidden field, I suggest we do that change.

yakky:
Fix #485

### DIFF
--- a/filer/fields/folder.py
+++ b/filer/fields/folder.py
@@ -18,7 +18,6 @@ from filer.settings import FILER_STATICMEDIA_PREFIX
 class AdminFolderWidget(ForeignKeyRawIdWidget):
     choices = None
     input_type = 'hidden'
-    is_hidden = True
 
     def render(self, name, value, attrs=None):
         obj = self.obj_for_value(value)


yakky:
@BertrandBordage would you double check this, please?

BertrandBordage:
What do you mean, double check this?
I already tried to remove `AdminFolderWidget.is_hidden = True` (for [cmsplugin-nivoslider#9](https://bitbucket.org/bercab/cmsplugin-nivoslider/issue/9/cant-create-content-please-correct-the)), and it worked flawlessly.

So this pull request is good to go.

yakky:
Thanks!

PR #500, opened 2015-02-23
import_files refactoring

skirsdeda:
This includes a rewrite of import_files which now has these features:
- Import from any Django storage
- Use standard File class detection (matches_file_type)
- Make import_files repeatable by adding duplicate file detection and user action selection. Now it is possible to import the same path to the same filer target folder by selecting --duplicate-action=update and only changed files will be overwriten, no duplicates will be created.

The big thing is size/image size/sha calculation optimization which turned out to be somewhat complicated. Now these file attributes are only recalculated from storage when file's url changes (the only proper way to detect if file contents were changed that I found). This could be improved though:
- File.sha1 could be renamed to _sha1, then get_sha() could be renamed to sha1 and become a property (just like size and image_size)
- get_file_id() could be improved and if we had a model field for this, database hit in data_changed() could be removed and FileQuerySet.iterator override wouldn't be necessary

### DIFF
--- a/filer/management/commands/import_files.py
+++ b/filer/management/commands/import_files.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, unicode_literals
 
+import inspect
 import os
 from collections import namedtuple
 from optparse import make_option
@@ -60,13 +61,14 @@ class FileImporter(object):
             if not qs.filter(file=obj.file.name, is_public=obj.is_public).exists():
                 obj.file.delete(False)
 
+        basename = os.path.basename(file_obj.name)
         # find the file type
         for filer_class in filer_settings.FILER_FILE_MODELS:
             FileSubClass = load_object(filer_class)
-            # TODO: What if there are more than one that qualify?
-            if FileSubClass.matches_file_type(file_obj.name, None, None):
+            #TODO: What if there are more than one that qualify?
+            if FileSubClass.matches_file_type(basename, None, None):
                 obj = FileSubClass(
-                        original_filename=os.path.basename(file_obj.name),
+                        original_filename=basename,
                         file=file_obj,
                         folder=folder,
                         is_public=filer_settings.FILER_IS_PUBLIC_DEFAULT)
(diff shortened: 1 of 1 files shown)


skirsdeda:
So apparently **FolderAdmin.resize_images** stopped working, because it sets **image.file.file** directly and file name doesn't change after that. Since it is very difficult to tell apart **file.file** set through **\__init\__** from file.file set by other code (which actually changes content), I would suggest rewriting **resize_images** so that file name changes after resize.

yakky:
@skirsdeda could you check the failure in travis?

skirsdeda:
My comment above (https://github.com/stefanfoulis/django-filer/pull/500#issuecomment-75755696) is about failure in travis. Is the proposed solution good?

yakky:
@skirsdeda sorry for not getting back sooner.
Go ahead with the proposed change

stefanfoulis:
@skirsdeda sorry for not merging this sooner. If you rebase we can merge this.

skirsdeda:
@stefanfoulis Good to hear. I'll get back to this then. I think size/image size/sha calculation part still needs some work.

mkoistinen:
@skirsdeda What's the status on this? We're preparing a release for soonish, will we be able to merge this into it?

skirsdeda:
@mkoistinen What is the timeframe (roughly) for 1.3.0?

skirsdeda:
I'll remove size/sha1 recalculation stuff from this PR and leave only import_files refactoring some time soonish.

jccode:
What's the status on this? Why not merge this PR? I am facing this issue as well, `import_files` command is not compatible with Django 1.10 , please refer to the issue #966 . Thanks

skirsdeda:
@jccode this has been finished more than 2 years ago :) But nobody cares to merge it.

jdcookie:
So this still seems to be an issue. Trying to use Django 1.11 and can't because of this issue in the import_files command. Is it possible to revisit this?

yakky:
@skirsdeda now that we are working towards 1.4, any chance you can rebase this on top of current develop?
Some cleanup has already been addressed in #1058, you can thus concentrate on the actual features

PR #502, opened 2015-03-03
Always use RequestContext instead of just Context

j00bar:

PR #503, opened 2015-03-09
dismiss_popup.html should be rendered with a RequestContext

gavinwahl:
It inherits from admin/base_site.html, which can request certain context variables. Without a request context, it won't have what it needs.


gavinwahl:
An alternative solution would be to not have dismiss_popup.html inherit from anything, because all it does is call some javascript to immediately close the window.

yakky:
Duplicate of #502

PR #523, opened 2015-05-04
Fix error in migrations caused by FilerFileField.

rockymeza:
FilerFileField (and friends) were passing direct references to models in
for the to parameter of ForeignKey. This worked previously with south
migrations, but seems not to work very well with migrations in Django
1.8.

Please see https://github.com/fusionbox/filer-migration-bug-example for
more information.

### DIFF
--- a/filer/fields/file.py
+++ b/filer/fields/file.py
@@ -10,6 +10,7 @@ from django.template.loader import render_to_string
 from django.utils.safestring import mark_safe
 
 from filer.utils.compatibility import truncate_words
+from filer.utils.model_label import get_model_label
 from filer.models import File
 from filer import settings as filer_settings
 
@@ -113,18 +114,15 @@ class FilerFileField(models.ForeignKey):
 
     def __init__(self, **kwargs):
         # We hard-code the `to` argument for ForeignKey.__init__
+        dfl = get_model_label(self.default_model_class)
         if "to" in kwargs.keys():  # pragma: no cover
-            old_to = kwargs.pop("to")
-            dfl = "%s.%s" % (
-                    self.default_model_class._meta.app_label,
-                    self.default_model_class.__name__
-            )
+            old_to = get_model_label(kwargs.pop("to"))
             if old_to != dfl:
                 msg = "%s can only be a ForeignKey to %s; %s passed" % (
-                    self.__class__.__name__, self.default_model_class.__name__, old_to
+                    self.__class__.__name__, dfl, old_to
                 )
                 warnings.warn(msg, SyntaxWarning)
-        kwargs['to'] = self.default_model_class
+        kwargs['to'] = dfl
         super(FilerFileField, self).__init__(**kwargs)
 
     def formfield(self, **kwargs):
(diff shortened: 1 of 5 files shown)


gavinwahl:
This is required for django 1.8 compatibility.

filer/fields/file.py
```diff
@@ -109,19 +109,15 @@ def widget_attrs(self, widget):
...
-                    self.default_model_class.__name__
-            )
-            if old_to != dfl:
+            if old_to != self.default_model_class:
                 msg = "%s can only be a ForeignKey to %s; %s passed" % (
-                    self.__class__.__name__, self.default_model_class.__name__, old_to
+                    self.__class__.__name__, self.default_model_class, old_to
                 )
                 warnings.warn(msg, SyntaxWarning)
         kwargs['to'] = self.default_model_class
```
rockymeza: instead of doing this, I thing that perhaps we could convert self.default_model_class to the '{app_label}.{class_name}' string in here, thus keeping backwards compatibility.

yakky:
@rockymeza this is failing when using custom model for Image. Could you have a look at this?
I'm eager to merge this

rockymeza:
I was working on it with my co-worker, and we got it working on everything
except Django 1.7 with the custom image model. But then I had to work on
some client stuff.

I'll try to take a look at it soon.

Thanks,
-rocky
2015年5月22日 下午3:53于 "Iacopo Spalletti" <notifications@github.com>写道：

acatton:
@yakky Unless I'm mistaking (which I could be) this is an Error in django-filer which was hidden before fixing this bug. Here's what happens:

* [`test_utils.custom_image.models` is importing `filer.models.abstract`](https://github.com/stefanfoulis/django-filer/blob/bb0bfa4bd9dccd18c05f01f4da2ef9e7ed1a8858/filer/test_utils/custom_image/models.py#L4)
* So this imports `filer`, `filer.models` and `filer.models.abstract`
* When importing `filer.models` it imports `*` from `filer.models.imagesmodels`, [`imagesmodels.py` which calls `load_object(filer_settings.FILER_IMAGE_MODEL)`](https://github.com/stefanfoulis/django-filer/blob/bb0bfa4bd9dccd18c05f01f4da2ef9e7ed1a8858/filer/models/imagemodels.py#L60)
* `load_object(filer_settings.FILER_IMAGE_MODEL)` is importing `test_utils.custom_images.models` which is what is being loaded in the first place. Therefore it raises an `ImportError`. (But django catches it while loading the app registry)

I have no idea how to fix it, but this import loop has to have happened before.

HomemBravo:
I had the same problem and did rockymeza workaround. On the migration that uses filer, instead of using filer.Image used filer.File

stefanfoulis:
I'd love to merge this, but we'll have to find a solution that works in Django 1.7 as well first.

yakky:
@stefanfoulis @rockymeza @HomemBravo  have a look at #559 which fixes this PR on Django 1.7
@acatton a refactoring of custom image models is under way to overcome this

PR #524, opened 2015-05-07
Added scope-wrapper function to use django.jQuery

strindhaug:
This makes filer work in admin modules that does not define $ or jQuery in the global scope.
(All the other scripts already use django.jQuery)


yakky:
Thanks @strindhaug

alexcustos:
One more fix needed to make the focal_point.js work correctly:
```
-    jQuery(document).ready(function() {
+    $(document).ready(function() {
```

PR #535, opened 2015-05-18
 	AttributeError: 'FileAdmin' object has no attribute 'queryset'

bashu:


yakky:
At least one ``self.queryset`` still need to be fixed https://github.com/stefanfoulis/django-filer/blob/develop/filer/admin/fileadmin.py#L100

yakky:
Thanks @bashu !

PR #536, opened 2015-05-21
Fix bad HTML - Remove extra quote character

vstoykov:
There is an extra quote character which will make the HTML invalid. Also brakes syntax highlighting of the file (because of invalid syntax)


yakky:
Nice!

PR #537, opened 2015-05-21
Show Uploaded at date and time in folder listing

vstoykov:
I'm not sure if you will like it but probably is good idea to show Uploaded at date in folder listing.

What do you think?


yakky:
I'm 0+ on this. I'm not 100% sure this is an information needed in all use cases
Ping @stefanfoulis

stefanfoulis:
I think we should keep the amount of information in the list view limited. I'd actually rather reduce it than add more.

PR #540, opened 2015-05-28
allow Django>=1.7 to find migrations automatically

stefanfoulis:
Puts south and django 1.7 migrations into the recommended folder. No need anymore to define      

```python
MIGRATION_MODULES = {
    'filer': 'filer.migrations_django',
}
```
in django 1.7. And with South>=1.0, south will also automatically pick up migrations from ``filer.south_migrations``.

I know the readme previously stated, that we'd change this in django-filer 1.0. But since South 1.0 has been release, there really is no reason to make life hard with the old setup.

I'd like to release this as django-filer 0.9.10.


dursk:
Are we sure this fix actually worked? I'm running `0.9.11`, and couldn't get `Django 1.7` to pick up the migrations until adding ```MIGRATION_MODULES = {
    'filer': 'filer.migrations_django'
}``` to my settings file.

PR #544, opened 2015-06-15
swap jQuery with $, as jQuery seems undefined

benzkji:
this works for me. if jQuery is used, I get a 

````
focal_point.js:80 Uncaught TypeError: undefined is not a function
````


yakky:
Fixed by #546 , sorry I spot that one before :)

benzkji:
argh. too late :/ glad  this works now. any timeframe for another release? dont like pip -e ... ;-)

yakky:
A few hours

benzkji:
thank you sir!

PR #546, opened 2015-06-17
Fix issue #545

### LINKED ISSUE #545: focal_point.js error
Going into Change Image options I am unable to use the focal point target and the following error is output to the console in Firefox: `TypeError: jQuery is not a function focal_point.js:80:1`

Same issue and similar errors occur in Chrome and Internet Explorer.

Changing line 80 in focal_point.js from

```
jQuery(document).ready(function() {
```

to

```
$(document).ready(function() {
```

seems to fix the problem

louis-haddrell:
Fix #545

### DIFF
--- a/filer/static/filer/js/focal_point.js
+++ b/filer/static/filer/js/focal_point.js
@@ -77,7 +77,7 @@ $(function(){
 		add(x, y);
 		image_loaded = true
 	}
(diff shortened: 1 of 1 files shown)


yakky:
Thanks @louis-haddrell !

Issue #515, opened 2015-04-15
NoReverseMatch when deleting an image with a custom model 


BertrandBordage:
I’m using a custom Image model, but the admin tries to reverse `filer_image_changelist` (see [filer/admin/fileadmin.py#L114](https://github.com/stefanfoulis/django-filer/blob/20bcc4c905898b4facd9fb2642f40d314eff4302/filer/admin/fileadmin.py#L114)). Of course it fails because filer.Image is not a registered model.

PR #549, opened 2015-06-23
fix for issue #515

svleeuwen:


yakky:
@svleeuwen thanks!

svleeuwen:
@yakky yw, small note: I saw that in https://github.com/stefanfoulis/django-filer/blob/develop/filer/admin/permissions.py#L33 there's a Django 1.7 specific check for getting the model_name. Might this be required here as well?

PR #557, opened 2015-07-13
Fix unsupported backend error when using s3 in django-storages

kriwil:
the storages backend doesn't implement path() as it should be, but filer requires that.


stefanfoulis:
awesome, thanks!

PR #558, opened 2015-07-13
You can set an app_label on your custom image model. Account for this…

svleeuwen:
… in image admin changelist url.

suggestion by @azuax


stefanfoulis:
thanks!

PR #559, opened 2015-07-16
Fix django18

yakky:
Fix #523 by adding migration dependecy


yakky:
@stefanfoulis Are you ok with this?

boydjohnson:
We are seeing this same issue (`AttributeError: 'NoneType' object has no attribute '_meta'`) when using any custom plugin models that use `cms.models.fields.PageField`.  This only occurs when we upgraded to the latest Django 1.8.x version. It doesn't appear that divio/django-cms#4275 is fixed in our case.  We have a patch available and will re-open that issue.  Just wanted to comment here since this was the first place where we thought the issue was located and discovered it was not the case.

PR #561, opened 2015-07-16
Remove unused 'thumbnail' template tags to avoid conflicts - fixes #560

### LINKED ISSUE #560: Suggesting to remove unused 'thumbnail' template tags to avoid conflicts
We are currently in a dilemma. We used ``easy_thumbnails`` and we have to use ``sorl.thumbnail`` as well. These two libraries work both a bit differently and are used in a lot of 3rd party apps. The main problem with these two apps is, that they both come with a ``"thumbnail"`` template tag.
Tom Christie suggested [on stackoverflow](http://stackoverflow.com/a/8183735/485361) to create some kind of proxy package, but this does not fix the problem in the global scope. 
A better approach, basically how to use ``django-smart-load-tag``, is described [here](http://timmyomahony.com/blog/using-sorl-thumbnail-and-easy-thumbnails-same-template/). We use this in our templates and plugins/apps.

It seems others (eg. #45) have the same conflicting issues. So, what does this all mean for django-filer? 

The ``thumbnail`` tag is loaded in two files, [here](https://github.com/stefanfoulis/django-filer/blob/develop/filer/templates/admin/filer/tools/clipboard/clipboard.html#L1) and [here](https://github.com/stefanfoulis/django-filer/blob/develop/filer/templates/admin/filer/tools/clipboard/clipboard_item_rows.html#L1). As far as I can see, the tag is loaded but not used so just remove them. If it has to be re-added in the furure, I suggest to go with ``django-smart-load-tag``.

philippeowagner:
BTW: There is a feature branch ready to merge django-smart-load-tag into filer on my side. 

Fix #560

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,3 +8,4 @@
 * Martin Brochhaus <mbrochh@gmail.com> https://github.com/mbrochh
 * Ioan Alexandru Cucu https://github.com/kux
 * Tim Graham https://github.com/timgraham
+* Philippe O. Wagner https://github.com/philippeowagner
--- a/filer/templates/admin/filer/tools/clipboard/clipboard.html
+++ b/filer/templates/admin/filer/tools/clipboard/clipboard.html
@@ -1,4 +1,4 @@
-{% load thumbnail i18n %}
+{% load i18n %}
 {% load url from future %}
 
 {% for clipboard in user.filer_clipboards.all %}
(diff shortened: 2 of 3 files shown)


yakky:
Failures in Django 1.4-1.5 depends on the djangocms-helper version.
This is fixed in another PR, merging this

yakky:
Thanks @philippeowagner !

PR #564, opened 2015-07-23
FolderAdmin uses base admin context in views

gaetano-guerriero:
Views in admin site use a context which includes properties like
"site_head" and "site_url". This properties are used by base admin
templates.
Previously, FolderAdmin was creating a new context in his views
functions, not including default properties.
Now, FolderAdmin starts from the base context of every admin view.
This uses function each_context which is available in django>=1.7, so
the new behaviour is only from that version upwards.


yakky:
@gaetano-guerriero Nice change! Please check python 2.6 compatibility

yakky:
@gaetano-guerriero thanks!

Issue #562, opened 2015-07-20
thumbnail background on preview


darioalbanesi:
When using images with a white logo on a transparent background, it is not possible to recognise the image on the filer thumbnail preview. A checkerboard background would solve the problem.

![Checkerboard Background](http://informationworker.ru/snagit10.en/images/checkerboard.png)

strindhaug:
I had the same problem and I've created a fix for this (noticed this issue when I was about to create the commit). 

Since I often find that the classical pattern with fairly large squares as in your example a bit distracting (it makes transparent gray images look jagged), I used a fairly light checkerboard pattern with small squares (4x4 px squares). To be able to see light transparent images properly it changes to a black and dark gray pattern when hovering the cursor.

lory87:
gentle ping @strindhaug Is this issue fixed for you in current filer version 1.2.0 so we can close?

strindhaug:
@lory87 sure!  :+1:

PR #566, opened 2015-07-29
Added checkerboard tiles as background to thumbnails. Fix for issue #562

strindhaug:
Thumbnails in the directory table and the preview image now has a light checkerboard pattern background (that turns dark on mouseover) to make both light and dark transparent images easier to see.

Uses two small gif's embedded in the css using data urls so that it doesn't add additional http requests.

### DIFF
--- a/filer/static/filer/css/admin_style.css
+++ b/filer/static/filer/css/admin_style.css
@@ -3,6 +3,16 @@
 #image_container { position: relative; } 
 #image_container img, #image_container svg { margin: 0px; position: absolute; top: 0; left: 0;}
 
+/* Tile pattern as background for transparent images: */
+.transpTiling {
+    /* light tiles */
+    background-image: url(data:image/gif;base64,R0lGODlhCAAIAKECAOPj4/z8/P///////yH5BAEKAAIALAAAAAAIAAgAAAINhBEZh8q6DoTPSWvoKQA7);
+    outline: 1px solid #E3E3E3;
+}
+.transpTiling:hover {
+    /* dark tiles */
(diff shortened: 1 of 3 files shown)


strindhaug:
The css class "transpTiling" can probably safely be added to any other wrapper element that contains the image (as long as the `filer/static/filer/css/admin_style.css` is included on relevant page/iframe); since it only adds the background image and a thin 1px outline it should not otherwise interfere much with the layout.

In this commit I've only added it to the thumbnail in the directory table, and the thumbnail where you can set the subject location.

yakky:
Fix #452 
@darioalbanesi, could you check if this solves the issue for you?

stefanfoulis:
@strindhaug I love the simplicity of including the image inline in the css :+1:

strindhaug:
In case you don't want to install a bleeding edge version to test this patch you can copy these files from the patch:
`filer/static/filer/css/admin_style.css`
`filer/templates/admin/filer/folder/directory_table.html`
`filer/templates/admin/filer/image/change_form.html`

to your main django app static and template directories. In my app that would be:

`myapp/static/filer/css/admin_style.css`
`myapp/templates/admin/filer/folder/directory_table.html`
`myapp/templates/admin/filer/image/change_form.html`

(Replace `myapp` with the name of your main app. You may need to make the directories)
This will override the templates and styles from the filer-plugin with the changed ones.
(If you've already made your own overrides of some of these files, you can back them up and replace them or manually merge the files)

Im using django-filer 0.9.11 in my app and (besides my changes) there is almost no difference between this version and the patched version of these files. (The only other difference is a minor change in formatting of number of files but this seems to be compatible anyway).

darioalbanesi:
@yakky veery nice! lgtm!

yakky:
@strindhaug Thanks for this, merging

PR #571, opened 2015-08-07
[Feature/FIX] FolderAdmin.search_fields now is used

belimawr:
FolderAdmin is ignoring FolderAdmin.search_fields (see [FolderAdmin.filter_folder](https://github.com/stefanfoulis/django-filer/blob/develop/filer/admin/folderadmin.py#L421-L427) and [FolderAdmin](https://github.com/stefanfoulis/django-filer/blob/develop/filer/admin/folderadmin.py#L60-L67))

It would be great to use ``FolderAdmin.search_fields``, so one can just extend ``FolderAdmin´´ to change the search behavior as described in [Django's docs](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields) . E.g.:
```python
class MyFolderAdmin(FolderAdmin):
    search_fields = ['=filed1', '^field2']

admin.site.unregister(Folder)
admin.site.register(Folder, MyFolderAdmin)
``` 

So I did it. :)

I also believe that ``FolderAdmin.owner_search_fields`` could use the same logic. However as it uses a wee different rule, I rather make it a different feature/pull request.

### DIFF
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -61,7 +61,7 @@ class FolderAdmin(PrimitivePermissionAwareModelAdmin):
     exclude = ('parent',)
     list_per_page = 20
     list_filter = ('owner',)
-    search_fields = ['name', 'files__name']
+    search_fields = ['name__icontains', ]
     raw_id_fields = ('owner',)
     save_as = True  # see ImageAdmin
     actions = ['move_to_clipboard', 'files_set_public', 'files_set_private',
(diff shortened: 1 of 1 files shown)


yakky:
Please rebase on current develop

belimawr:
I'm making another pull request, because my local branches are a bit messy.

yakky:
@belimawr thanks

Issue #573, opened 2015-08-11
Duplicated folder destinations in move files/folders


o-zander:
On the move files/folders page accessed via the action in folder admin there is an issue with the destination field. It should display a tree like structure of all available destination folders. But it shows all folder dozen of times. Why? The problem is here https://github.com/stefanfoulis/django-filer/blob/30bc8c84ad2d365e6ba80232aad3aabd7df7c28c/filer/admin/folderadmin.py#L809 When permissions are activated FolderRoot.children returns ALL folders not just the root ones. So the recursive function visits each and every folder multiple times by themself and their parents. What has to be done? The initial folder queryset has to be reduced to root folders. I don't know why the FolderRoot class returns all folders when permissions are activated so I have not created a fix for the issue. I should be decided whether the issue is resolved by changing the FolderRoot behaviour or by just filtering the queryset on this specific line.

I'm currently using the latest version (0.9.12) but I examined the changes on this topic and for me it seems this issue exists forever and no one noticed ;)

mkoistinen:
Fixed by #577.

lory87:
gentle ping @o-zander Is this issue fixed for you as well and can we close it?

o-zander:
Yes, since #577 is merged this issue is fixed.

PR #577, opened 2015-08-20
Fix for issue #573: duplicate destination folders on move/copy action…

o-zander:
… when permissions are enabled


mkoistinen:
LGTM, thank you!

PR #579, opened 2015-08-27
Fix django comparison

acatton:
* If you installed Django==1.8.0, django.get_version() > '1.8' was
    False
  * If you would to install Django==1.10, django.get_version() > '1.8'
    would be False
  * Strings are not meant to be compared like numbers

### DIFF
--- a/filer/admin/tools.py
+++ b/filer/admin/tools.py
@@ -1,18 +1,19 @@
 #-*- coding: utf-8 -*-
 
-import django
 from django.core.exceptions import PermissionDenied
 
+from filer.utils.compatibility import DJANGO_1_7, DJANGO_1_6
 
-if django.get_version() > '1.8':
+
+if DJANGO_1_6:
     def admin_each_context(admin_site, request):
-        return admin_site.each_context(request)
-elif django.get_version() > '1.7':
+        return {}
+elif DJANGO_1_7:
     def admin_each_context(admin_site, request):
         return admin_site.each_context()
 else:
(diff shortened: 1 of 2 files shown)


yakky:
@acatton thanks. My bad: I haven't checked that before merging.
An even better solution is something like:

    from filer.utils.compatibility import DJANGO_1_6
    if DJANGO_1_6:
        def admin_each_context(admin_site, request):
            return {}
    else:
        def admin_each_context(admin_site, request):
        return admin_site.each_context()

Do you mind updatng the PR?

acatton:
@yakky I can't just do a if else, because Django 1.8 actually requires to pass the `request` to `each_context()`.

Patch updated. I also removed generating a string from `django.VERSION` (which is what `.get_version()` does and reparsing it to a tuple) making the code simpler.

Let me know if there is any problem.

yakky:
@acatton sorry for overlooking the request parameter. order must be actually reversed. Something like


    from filer.utils.compatibility import DJANGO_1_6
    if DJANGO_1_6:
        def admin_each_context(admin_site, request):
            return {}
    elif DJANGO_1_7:
        def admin_each_context(admin_site, request):
            return admin_site.each_context()
    else:
        def admin_each_context(admin_site, request):
            return admin_site.each_context(request)

As the DJANGO_* variables means **equals or lower** so they must be evaluated from lowest to highest

acatton:
@yakky, done. Thanks for the review.

yakky:
@acatton thanks, great

PR #580, opened 2015-09-01
Remove use of deprecated Django functions

gavinwahl:
All this stuff has deprecated for a long time, and is getting removed in Django 1.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ install:
   - ./.travis_setup
 
 env:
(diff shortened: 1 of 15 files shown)


mattbriancon:
Can we get this merged? Part of the decrease in test coverage is due to changes that have already made it develop; if we can get this branch rebased it should be pass coveralls.

yakky:
@gavinwahl this needs to be rebased on top of current develop

stefanfoulis:
I think these things have since been updated on develop. thanks!

PR #586, opened 2015-09-09
Give user hint about why the clipboard arrow is grey and disabled

foobacca:
by adding a title attribute (which will show hover text).  Otherwise
users may think the clipboard is broken.


mkoistinen:
LGTM. Can anyone confirm that permissions failure is the only reason that the arrow would be grey?

stefanfoulis:
since the visual revamp this button no longer exists.

PR #591, opened 2015-09-18
URLs that stay the same when a file is replaced. Fixes #478

### LINKED ISSUE #478: URLs of replaced files
Say you have a file in the filer, and a filer CMS plugin that links to it. 

The URL used by that plugin will be disseminated by users.

Now we update the file, and use Advanced > change to upload a new version for the same Filer file.

Now visitors who come to get the file via the plugin will get the new version, but ones who are using a direct URL to the file will hit the old version

I think that this only affects files with URLs generated with `randomize()` - date-based URLs don't change.

jvamvas:
Sometimes users share the direct URL of public files. But whenever they update the file in the "advanced" panel on the file's admin page, the direct URL changes as well.

Perhaps django-filer could include a canonical URL that redirects to the latest version of a public file and that can be shared safely:

screen shot 2015-09-18 at 17 04 17

If they wish to use canonical URLs, users need to include filer's URLConf in their project's `urls.py`. If they don't configure their project that way, or if they create a non-public file, the filer admin does not display a canonical URL.


filer/models/filemodels.py
```diff
@@ -241,6 +241,19 @@ def url(self):
         return r
 
     @property
+    def canonical_url(self):
+        url = ''
+        if self.file and self.is_public:
+            try:
+                url = urlresolvers.reverse('canonical', kwargs={
+                    'uploaded_at': self.uploaded_at.strftime('%s'),
```
jvamvas: We can use the `File.uploaded_at` field, which is non-editable, to make the public file's URL a little bit more hard to guess.

filer/test_utils/templates/404.html
```diff
@@ -0,0 +1 @@
+404
```
jvamvas: This is required to make the unit test Django 1.4 compatible.

test_settings.py
```diff
@@ -50,7 +52,8 @@
         'easy_thumbnails.processors.filters',
     ),
     'FILE_UPLOAD_TEMP_DIR': mkdtemp(),
-    'FILER_IMAGE_MODEL': False
+    'FILER_IMAGE_MODEL': False,
+    'TEMPLATE_DIRS': (os.path.join(BASE_DIR, 'django-filer', 'filer', 'test_utils', 'templates'),),
```
jvamvas: This is required to make the unit test Django 1.4 compatible.

yakky:
@jvamvas really nice
:+1: from me

filer/urls.py
```diff
@@ -0,0 +1,8 @@
+#-*- coding: utf-8 -*-
+from django.conf.urls import url
+
+from filer import views
+
+urlpatterns = [
+    url(r'canonical/(?P<uploaded_at>[0-9]+)/(?P<file_id>[0-9]+)/$', views.canonical, name='canonical'),
```
yakky: How about making the ``canonical`` part of the path configurable?

jvamvas: Done with 19b7815.

yakky:
@stefanfoulis @mkoistinen is this good for you?
If so I'll resolve the conflicts and merge it myself

stefanfoulis:
@yakky :+1: from me :-)

yakky:
@jvamvas merged. Thanks a lot for tour great work

PR #593, opened 2015-09-28
Avoid nested forms by inserting the clipboard after the changelist form.

EnTeQuAk:
Since https://github.com/stefanfoulis/django-filer/commit/f58cb6ade060560b6a8ac75484b719effcc27b20 the folder clipboard gets inserted directly into the changelist and thus there are nested forms which is not allowed and not supported (at least here my latest chrome is failing hard by simply not showing the `paste clipboard` form)

Rendering the clipboard after the changelist-form get's closed doesn't change visuals (imho) but fixes nested forms.


yakky:
LGTM
Ping @stefanfoulis

EnTeQuAk:
I support https://github.com/stefanfoulis/django-filer/commit/d9a40bddb86d46e2d86539e39d6af423c30f8289 does this then...

PR #597, opened 2015-10-11
django.utils.importlib deprecated since django 1.7

javierwilson:
See https://docs.djangoproject.com/en/1.8/releases/1.7/#django-utils-dictconfig-django-utils-importlib
django.utils.dictconfig and django.utils.importlib were copies of respectively logging.config and importlib provided for Python versions prior to 2.7. They have been deprecated.

### DIFF
--- a/filer/utils/loader.py
+++ b/filer/utils/loader.py
@@ -9,7 +9,7 @@ local changes:
 
 """
 from django.utils import six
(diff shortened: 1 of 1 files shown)


mkoistinen:
Correct, but this package supports Django 1.5+.

PR #600, opened 2015-11-04
Make user unique for clipboard

julianandrews:
This prevents a race condition in `get_user_clipboard` which can result in:

```
MultipleObjectsReturned: get() returned more than one Clipboard -- it returned 2!
```

errors. According to the [django docs](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#get-or-create) `get_or_create` is only atomic if uniqueness is enforced at the database level.


julianandrews:
I thought about line-wrapping for PEP008 compliance, but since the line was already overlong I left it alone. Let me know if you'd like me to fix that up.

mkoistinen:
@julianandrews This looks like a sensible change. I'd like to merge it, but I think it will need to have refreshed migrations. Can you do this, or shall I create a new PR with this change in?

julianandrews:
@mkoistinen I've updated the PR. Let me know if you need any other changes.

mkoistinen:
@julianandrews Thank you! I think we also require South migrations too though...

julianandrews:
@mkoistinen I'm not really setup to generate south migrations. I didn't realize filer still supported south! Do you maybe have a project that can be used to generate those?

vstoykov:
This branch need to be rebased

JeremyGibson:
Just wanted you guys to know that my company is working on a project, and we ran into this race condition on one of our projects.

PR #606, opened 2015-11-13
Fixes bad static path

ipmb:


yakky:
Thanks for the fix @ipmb

PR #608, opened 2015-11-20
adds hotfix for django 1.8

czpython:
Basically django 1.8 no longer needs the onclick() action so if rendered then it causes the list endpoint to be called twice and create a race condition in multi-thread environments.


mkoistinen:
LGTM

FinalAngel:
looks good, does that include the admin js?

FinalAngel:
LGTM

fenginx:
@czpython Could you please explain that?  I actually had issues with the 1.8 hack in template and had to remove that 1.8 check logic in order to get it work again with django-widgy.  And I didn't observe the race condition as mentioned.

PR #636, opened 2015-12-14
CMS-913: Image detail view has to much empty space on top

Flight:
Fixed:
Go to: Adminstration -> Django filer -> Folders
Click on a uploaded image, you see a lot of empty space on top (see screenshot)
This should behave the same way as when you open image detail view from a image plugin.


Brankub:
:+1:

PR #637, opened 2015-12-14
CMS-916: optimise scrolling experience on admin filer

Flight:
Fixing fixed position of breadcrumbs while scrolling


dhorelik:
lgtm

PR #640, opened 2015-12-17
Feature/drag and drop widget

Flight:
This PR implements drag and drop for the widgets in the structure mode.

Please **do not review dropzone.scss**.

**JS is also outdated**, just tried to leave the old styling (because it can be used by some other plugins).


filer/private/sass/layout/_main.scss
```diff
@@ -566,3 +566,46 @@ body {
...
+    overflow: hidden;
+    background: #FFF url('/static/filer/icons/nofile_48x48.png');
+    background-size: cover;
+    border: solid 1px #ddd;
+    border-radius: 3px;
+    margin-right: 10px;
+    display: inline-block;
+    vertical-align: top;
+    img[src=''],
+    img:not([src]) {
```
Brankub: Did you check if you can use this with oldest version of IE which you support?

Flight: yes, we also using math calc in the admin styles.

ruslanmalogulko:
:+1:

filer/private/sass/layout/_main.scss
Brankub: :+1:

PR #641, opened 2015-12-21
Pin mptt version

yakky:
django-mptt 0.8 has some compatibility issue with filer / django CMS 3.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
     install_requires=(
         'Django>=1.5',
         'easy-thumbnails>=1.0',
-        'django-mptt>=0.6',
+        'django-mptt>=0.6,<0.8',
         'django_polymorphic>=0.7',
         'Unidecode>=0.04',
     ),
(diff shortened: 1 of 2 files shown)


yakky:
@mkoistinen @FinalAngel I think this breaks any filer + CMS 3.0 installation which does not pin mptt version and we should consider a quick filer release to fix this

FinalAngel:
LGTM

setup.py
```diff
@@ -27,7 +27,7 @@ def read(fname):
     install_requires=(
         'Django>=1.5',
         'easy-thumbnails>=1.0',
-        'django-mptt>=0.6',
+        'django-mptt>=0.6,<0.8',
```
cyberstar: @yakky 0.8 release drops ``Django <1.8`` AND ``Python 2.6`` and I don't thinks that this is correct to limit  mptt version for python 2.7 for example. Maybe it is better to adjust ``README``'s and limit mptt version only for ``Python<2.7``?
something like https://github.com/aldryn/aldryn-events/blob/master/setup.py#L35

yakky: You can't express versioned dependency on Django, and until filer supports django<<1.8 we can't include mptt>=0.8 in dependencies.
The major issues is with tests, in which you have to pin mptt in every Django 1.6/1.7 environment, which seems a huge burden.

cyberstar: can we switch from travis envs to running tox commands the same way as on ``aldryn-events`` or ``aldryn-newsblog`` or ``aldryn-jobs``? I've adjusted ``tox`` setup here https://github.com/divio/django-filer/pull/643

yakky: The problem is all the other applications which declare dependency on filer, which must pin the mptt version whenever they depend on filer, which seems rather unconfortable

PR #642, opened 2015-12-21
Fix admin ajax upload url regex.

cyberstar:
at some point previous regex with ``\d``was broken for subfolders, something escaped the backslash...


intellectronica:
are you sure it should be a `*` and not a `+`? is it ok to not have any digits following?

mkoistinen:
I agree with @intellectronica, perhaps this should be `\d+`?

PR #643, opened 2015-12-21
Adjust tox setup with respect to mptt changes.

cyberstar:
Use general approach in requirements handling.


mikek:
Thanks, implemented in [1.0.6](https://github.com/divio/django-filer/releases/tag/1.0.6).

PR #645, opened 2015-12-21
Drag-n-drop to the table view

Flight:
This PR implements the drag-n-drop to the table view (including multi-upload and upload progress).


Flight:
About travis:
It's only fails on BE tests, FE are ok.
@cyberstar knows this

Brankub:
:+1:

PR #651, opened 2015-12-28
HotFix django-polymorphic requirements

suquant:
Current version not support django_polymorphic>=0.8

For example:

```
ImportError: cannot import name PolymorphicModel
```

PolymorphicModel class not exists in polymorphic module.
It need be imported from polymorphic.polymorphic_model import ...

This is a hotfix


mikek:
Thanks, fixed in [1.0.6](https://github.com/divio/django-filer/releases/tag/1.0.6).

PR #652, opened 2015-12-28
Newest django-polymorphic support

suquant:
Support django_polymorphic>=0.8,<0.9

In newest version of django_polymorphic deleted some back compatibilities, for example there was deleted "PolymorphicModel" in polymorphic/\__init\__.py


setup.py
```diff
@@ -25,11 +25,11 @@ def read(fname):
...
     install_requires=(
-        'Django>=1.5',
-        'easy-thumbnails>=1.0',
-        'django-mptt>=0.6',
-        'django_polymorphic>=0.7',
-        'Unidecode>=0.04',
+        'Django>=1.5,<1.9',
+        'easy-thumbnails>=1.0,<2.4',
+        'django-mptt>=0.6,<0.9',
+        'django_polymorphic>=0.7,<0.9',
```
dustinfarris: should be `<0.8`

dustinfarris:
This is trying to support both 0.7 and 0.8 at the same time?  Does not seem possible with the breaking change?

mikek:
Thanks bringing this up. 

We're not likely to pin django-mptt in setup.py though, as it really depends on a Django version. We'll probably pin it in tests and let the user decide.

We'll look into supporting both django-polymorphic versions, if that is possible. (Besides the changed import there are also problems with custom queryset methods.)

setup.py
suquant: This line https://github.com/divio/django-filer/pull/652/files#diff-c5ee97a47ce37023a9ab5393b05721bfR14

fixing support from 0.8 until 0.9 because there >=0.7,<0.9 and 0.9 not included

suquant:
@mikek When command

```
pip install --upgrade -r requirements.txt
```
success result depends on time this is not stable solution. 

In .travis.yml (https://github.com/divio/django-filer/blob/develop/.travis.yml) not contains tests fo django 1.9 but in setup.py requirements appears "Django>=1.5"

mikek:
@suquant I know =). We'll refactor travis setup.

suquant:
@mikek tnx :) :+1:

mikek:
@suquant btw, do you have any specific reason to pin `unidecode` and `easy-thumbnails`? Otherwise we can try to test/support a broader set of versions.

suquant:
@mikek In my opinion we need limit any library to specific "major" and "minor" number, by this way less chance get some bugs.

mikek:
Thanks, fixed in [1.0.6](https://github.com/divio/django-filer/releases/tag/1.0.6).

P.S.
I've tried to merge your PR initially, but had to redo from scratch in master.

PR #654, opened 2015-12-29
Pin django-polymorphic

mikek:
Pin django-polymorphic version to be below 0.8.

Also pins django-mptt in tox.

Note: Travis build is currently broken only because of django-mptt version being dependent on Django version. We'd better merge this quickly and do a version bump/release to fix issue for all the possible users.

The next steps: 

 * fix travis setup based on #643 
 * really support new django-polymorphic (broken because of imports and queryset methods proxying)


cyberstar:
LGTM

PR #664, opened 2016-01-05
changelist paginator styles and column height

lory87:

|BEFORE|AFTER|
|---|---|
|screen shot 2016-01-05 at 16 06 17|screen shot 2016-01-05 at 16 22 48|


Flight:
LGTM

PR #674, opened 2016-01-07
folder admin: add missing search results querysets to context

mikek:
search_form.html template makes use of these variables, but they were
not set in the context.


czpython:
:+1:

PR #681, opened 2016-01-08
Updated table drag-n-drop logic

Flight:
* now the file can be dropped everywhere
* "unfiled files" instead of "root" in popups
* old stuff removed


Brankub:
:+1:

PR #682, opened 2016-01-08
Move thumbnail option

### LINKED ISSUE #406: Move ThumbnailOption to filer
@stefanfoulis how about moving ThumbnailOption for cmsplugin_filer to filer? It really belongs here, IMHO, and in many case you can leverage it without the need of the whole cmsplugin.
An easy migration path is needed, though

romolox:
This pull request moves the ThumbnailOption model from 'cmsplugin-filer' to 'filer' so that it will be possible to use it without necessarly installing cmsplugin-filer.

When released cmspluginfiler will need to depend on a filer version with the thumbnailoption model. this pr eventually needs datamigrations to move existing thumbnailoption data from cmsplugin_filer to filer before dropping the old model. Or the "new" thumnailoption model can have a db_table attribute which maps over the legacy name and thus no datamigration is required.

Fix #406 
Linked to divio/cmsplugin-filer#208


yakky:
@stefanfoulis @mkoistinen any comments on this. I spoke with stefan a looong time ago about this (see linked issue)

stefanfoulis:
I think it would be a good idea to move this functionality to django-filer. The migration needs to be well thought out though. Off the top of my head, the following sequence could work:

1. Add the ThumbnailOption feature to django-filer (in a way that both cmsplugin-filer and this new version of django-filer can work side-by-side).
2. Release a new version of django-filer
3. Change cmsplugin-filer to use the new ThumbnailOption from django-filer. With a clean migration.
4. Release cmsplugin-filer with a dependency link >= the new version of django-filer

Important: django-filer does not know about and needs to work completely without cmsplugin-filer... so all that migration logic needs to be in cmsplugin-filer.

romolox:
@stefanfoulis @yakky I finished working on both filer and the cms-plugin-filer-image https://github.com/divio/cmsplugin-filer/pull/208/files
I've generated migrations both with south and django migrations.

yakky:
@stefanfoulis if you're ok with the migrations setup, we will fix the tests and will complete the migration

stefanfoulis:
This PR :+1: 
There are still issues with the south migration in the cmsplugin-filer south migrations

yakky:
@stefanfoulis this is complete. The failing  continuous-integration/travis-ci/push  test is because I mistakenly pushed this branch to this repo instead of the forked one and them removed the spurious branch

stefanfoulis:
:+1: :-)

PR #684, opened 2016-01-11
fix focal point position

lory87:

|BEFORE|AFTER|
|---|---|
|screen shot 2016-01-11 at 11 44 07|screen shot 2016-01-11 at 11 44 17|


sergii-kulakovskyi:
:+1:

PR #685, opened 2016-01-11
Adds max uploader connections (for old DB) + fixes small table width issue

Flight:
* Workaround for SQLite problems on simultaneous file uploads
 * Also fixes table width for non-CMS admin templates


lory87:
frontend :+1:

filer/settings.py
```diff
@@ -215,6 +215,16 @@ def update_server_settings(settings, defaults, s, t):
 FILER_PRIVATEMEDIA_SERVER = load_object(FILER_SERVERS['private']['main']['ENGINE'])(**FILER_SERVERS['private']['main']['OPTIONS'])
 FILER_PRIVATEMEDIA_THUMBNAIL_SERVER = load_object(FILER_SERVERS['private']['thumbnails']['ENGINE'])(**FILER_SERVERS['private']['thumbnails']['OPTIONS'])
 
+# By default limit number of simultaneous uploads if we are using SQLite
+FILER_NO_CONCURRENT_UPLOADS_ON_SQLITE = getattr(
```
czpython: I don't like the idea of a setting specifically targeted to sqlite3. Seems a bit overkill.

Maybe the behavior should be:

If sqlite3 then **as default** set connections to 1 else set to 3.
If FILER_UPLOADER_CONNECTIONS is set in settings then leave it alone.

Although as I'm writing this I can see it might be nightmarish for things like travis.. ahhh

czpython:
backend :+1:

PR #690, opened 2016-01-13
remove input clear on IE

lory87:

|BEFORE|AFTER|
|---|---|
|screen shot 2016-01-13 at 09 05 47|screen shot 2016-01-13 at 09 07 02|


vxsx:
:+1: 
we should probably do that for all inputs

lory87:
There are some problems on travis. @mikek is already on it. gulp ci is ok so I merge this pr

PR #695, opened 2016-01-13
Dropzone mobile view fix

Flight:
Dropzone fixes for cmsplugin-filer.


filer/static/filer/js/addons/dropzone.init.js
```diff
@@ -20,8 +20,8 @@
         var mobileClass = 'dropzone-mobile';
         var objectAttachedClass = 'js-object-attached';
         var minWidth = 500;
-        var checkMinWidth = function () {
-            this.toggleClass(mobileClass, this.width() < minWidth);
+        var checkMinWidth = function (element) {
```
vxsx: why not a media query?

Flight: because it can be in popup + it's not counting window width

vxsx: ah, it's more of an element query then, my bad :)

vxsx:
:+1:

PR #699, opened 2016-01-14
rename selectors to avoid conflicts

lory87:

To avoid conflicts if a user uses dropzone on his project as well I updated selectors.


vxsx:
:+1:  although it's more important to add them https://github.com/aldryn/aldryn-bootstrap3/pull/67 and in cmsplugin-filer (if it'll be updated to support dragndrop)

thanks!

lory87:
@vxsx on it now to change them on bootstrap as well.

PR #701, opened 2016-01-14
Handle bad file AJAX upload

mikek:
Handle bad files/filenames AJAX uploads.

This also reorders configured icon sizes and fetches available icons only.


crash843:
:+1:

PR #704, opened 2016-01-15
Fix documentation wrt tests and supported versions

mikek:
* use stricter Django verison in setup.py
 * remove broken support for test runner in setup.py 
 * fix tox arguments passing
 * update docs


crash843:
:

PR #706, opened 2016-01-15
Multi-theading fixes, focal point fixes, files max-size fixes, unit-t…

Flight:
…ests fixes

* multiupload fixes
* max file size added
* focal point fixes (when image is broken)
* unit-tests fixes


vxsx:
looks very good

only one question - can blank.png be smaller? like 1x1?

Flight:
@vxsx no, because I need the real img.naturalWidth

PR #708, opened 2016-01-18
Cleanup docs: merge most part of the README and wiki/contributing section

mikek:
We have lots of duplicate information in README, docs (RTD), GitHub wiki. This PR is an essential cleanup of such issues. We still have to move some relevant parts of the wiki to the docs before disabling it.


cyberstar:
LGTM, but maybe consider the prefered  markup as described here https://django-cms.readthedocs.org/en/develop/contributing/documentation.html#documentation-markup

mikek:
@cyberstar, sure. I've just followed the existing convention in the docs. We'll have to restructure them soon anyway.

PR #709, opened 2016-01-19
Convert FILER_ADMIN_ICON_SIZES elements to strings

mikek:
There is some code in the wild that depends on this being strings. Namely, django-cms plugin Filer.


czpython:
:+1:

PR #712, opened 2016-01-20
Ordering files by last modified after upload

jakob-o:
So you can find the just uploaded files faster.


stefanfoulis:
@lory87 could you please check if this pr still works with recent JS changes?

PR #713, opened 2016-01-22
Fix the bug that only the first rendered filefield gets Dropzone atta…

lfhong:
When there are multiple FileField in one admin page, the Dropzone feature is only activated for the first rendered field.


Flight:
FE LGTM

PR #735, opened 2016-02-11
Return of the self.file_ptr hack.

intellectronica:
Makes filer's File object not choke on cascade deletes of
objects referencing them.


intellectronica:
This obviously doesn't work, so back to the drawing board.

PR #747, opened 2016-02-23
updates search field size on load and on key enter

lory87:

- search field should be focused on load but small
- on keydown field should resize to bigger field same as on focus


filer/static/filer/js/base.js
```diff
@@ -59,9 +59,13 @@ Cl.mediator = new Mediator();
         // Focus on the search field on page load
         (function () {
             var filter = $('.js-filter-files');
+            var container = '.js-filter-files-container';
```
sergii-kulakovskyi: may be better to use `var containerSelector`

sergii-kulakovskyi:
:+1:

PR #758, opened 2016-03-01
updated drag and drop widget

lory87:

|BEFORE|AFTER|
|---|---|
|screen shot 2016-03-01 at 08 55 07|screen shot 2016-03-01 at 09 08 36|
|screen shot 2016-03-01 at 08 55 16 2|screen shot 2016-03-01 at 08 53 34 2|
|screen shot 2016-03-01 at 08 55 19|screen shot 2016-03-01 at 08 53 32|


filer/private/sass/settings/_custom.scss
```diff
@@ -47,6 +48,8 @@ $screen-mobile: 320px;
 $screen-tablet: 720px;
 $screen-desktop: 975px;
 
+$screen-filer: 810px;
```
vxsx: what is this value really?

lory87: I had to add a separate value for filer because height will break if I choose screen tablet.

vxsx: i was just thinking if we could come up with something that is a bit more descriptive than just `filer`

vxsx:
:+1:

filer/private/sass/settings/_custom.scss
lory87: I updated name to filer-widget

vxsx:
:+1:

PR #760, opened 2016-03-02
Fix wrong icon path in admin file widget

squiddy:
The path `filer/icon/nofile_48x48.png` doesn't exist, use the correct one.

This bug surfaced when using ManifestStaticFilesStorage in combination with djangocms-page-meta. Accessing a non existing file raised an exception, which broke the forms render method, therefore not rendering the form at all.


bh:
Thanks for this PR, I also have this issue.

stefanfoulis:
thanks!

PR #763, opened 2016-03-07
fix infinite template recursion in Django 1.6

stefanfoulis:
Django 1.6 does not support including the same template in itself, even
if a conditional would prevent infinite recursion at runtime.
See https://code.djangoproject.com/ticket/3544

This solution is also more efficient, as it only causes one query
instead of at least n+1, where n is the folder depth.

I also sneaked in a little bit of cleanup in this PR.


intellectronica:
:+1:

lory87:
:+1:

PR #767, opened 2016-03-07
adds order after upload

lory87:

- after uploading an image, uploaded image will be ordered by latest upload (first in row)
- after going away and back to this folder. uploaded image will be ordered by name.


FinalAngel:
LGTM

PR #779, opened 2016-03-14
fix folder delete view fails

### LINKED ISSUE #776: 500 error delete view with mixed objects
If the delete view has polymorphic results, django falls over polymorphic.

to reproduce:
add images and files to a folder and try to delete the folder

stefanfoulis:
fixes #776

### DIFF
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -38,7 +38,6 @@ from ..models import (
 from ..settings import FILER_PAGINATE_BY
 from ..thumbnail_processors import normalize_subject_location
 from ..utils.compatibility import (
-    LTE_DJANGO_1_6,
     capfirst,
     get_delete_permission,
     quote,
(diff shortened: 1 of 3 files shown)


filer/admin/folderadmin.py
```diff
@@ -178,17 +185,12 @@ def delete_view(self, request, object_id, extra_context=None):
...
-        admin_context = AdminContext(request)
-        if LTE_DJANGO_1_6:
-            extra_context = extra_context or {}
-            extra_context.update({'is_popup': admin_context.popup})
         if request.POST:
-            # Popup in pick mode. Call super delete view so the objects
-            # actually get deleted. All possible failures in delete_view cause
-            # exceptions, so it is safe to ignore the return value though.
-            super(FolderAdmin, self).delete_view(
-                request=request, object_id=object_id,
```
mkoistinen: You're probably doing it perfectly, but it doesn't seem right that you're not calling super() anymore.

stefanfoulis: yeah, it's a bit weird. But the alternative would have been copy-n-pasting the entire delete_view and replacing one line on it.

mkoistinen:
LGTM

filer/admin/patched/admin_utils.py
```diff
@@ -43,31 +48,103 @@ def format_callback(obj):
...
 
     protected = [format_callback(obj) for obj in collector.protected]
-
-    return to_delete, perms_needed, protected
+    model_count = {model._meta.verbose_name_plural: len(objs) for model, objs in collector.model_objs.items()}
+
+    return to_delete, model_count, perms_needed, protected
+
+
+class NestedObjects(Collector):
```
czpython: @stefanfoulis why not subclass the one from django admin?

stefanfoulis: Because I am using a copy of the Django 1.9 implementation of get_deleted_objects, and that only works with the matching 1.9 version of NestedObjects. When we drop Django<1.8 support this can probably be removed again.

PR #781, opened 2016-03-25
Fix renaming files without folders

mikek:
- Fix 'rename_files' action for 'unfiled' files.
- Add a few simple admin view tests.


cyberstar:
LGTM

PR #782, opened 2016-03-28
Image admin: subject location validation

mikek:
Rely on thumbnail subject location normalization method to validate
form field.

 - Make field non-nullable to make working with empty values easier
 - Adjust tests and add specific ones for subject location validation
 - Preserve subject location on image resize bulk action, adjust tests
 - Update Changelog


filer/models/abstract.py
```diff
@@ -36,8 +36,8 @@ class BaseImage(File):
     default_alt_text = models.CharField(_('default alt text'), max_length=255, blank=True, null=True)
     default_caption = models.CharField(_('default caption'), max_length=255, blank=True, null=True)
 
-    subject_location = models.CharField(_('subject location'), max_length=64, null=True, blank=True,
-                                        default=None)
+    subject_location = models.CharField(_('subject location'), max_length=64, blank=True,
```
mikek: Without this change there were a few tests failing because `model_to_dict()` had produced data with `subject_location=None` but form validation did not allow to save it (as expected). So, if there are no really good reasons to keep `Null=True` for a text field here, it will be removed.

cyberstar:
LGTM

stefanfoulis:
:+1: 
I'm not sure anymore, but there might be an admin action that allows batch resizing of images... it might be worth making sure the subject location is validated there too.

mikek:
@stefanfoulis, The validation was there, but the logic to preserve subject location was not really implemented. So, here it goes. Could you have a look?
@lory87, note that subject location code requires a bit of JS/style fixes anyway.

filer/admin/folderadmin.py
```diff
@@ -1169,8 +1169,21 @@ def _list_all_to_resize(self, request, files_queryset, folders_queryset):
...
+        # TODO: We could probably do even better, but this method knows nothing
+        # about actual thumbnailing algorithm details.
+        # It's better to reset subject location to the central point of the new
+        # image if any of these two conditions are met:
+        # - 'scale_and_crop_with_subject_location' thumbnail processor is
+        # disabled. Resizing would not respect subject location.
+        # - The image was cropped. The originally specified subject location
+        # is either near the new center (if subject location was respected) or
+        # could be outside of the new image.
+        if crop or 'filer.thumbnail_processors.scale_and_crop_with_subject_location' not in django_settings.THUMBNAIL_PROCESSORS:
```
stefanfoulis: Why check for the thumbnail processor setting? Without it the whole subject location machinery does not make sense, so why special case it here? If we go down that route we'd may as well hide the subject location UI in admin as well.
'If crop' should be enough here.

stefanfoulis:
Probably needs a data migration in case there are existing entries with None as subject location.

stefanfoulis:
Other than above comments :+1:
Awesome stuff!

mikek:
@stefanfoulis actually a data migration is not needed since there is a `default=''` specified for the field (that is enough for South/Django to do the conversion).

filer/admin/folderadmin.py
mikek: You're probably right. I've adjusted this condition and tests (just doing the same assertions for both original and subject-aware thumbnail processors now).

cyberstar:
LGTM

PR #786, opened 2016-03-31
hides related widget wrapper links

lory87:

|BEFORE|AFTER|
|---|---|
|screen shot 2016-03-31 at 10 43 04|screen shot 2016-03-31 at 10 51 07|


filer/private/sass/components/_drag-and-drop.scss
```diff
@@ -2,6 +2,9 @@
 // DRAG AND DROP
 
 form .form-row {
+    .related-widget-wrapper-link {
```
vxsx: this is very unspecific

vxsx: can it affect other fields in the form like this that actually require related-widget stuff?

Brankub:
:+1:

vxsx:
:+1:

PR #787, opened 2016-03-31
adds back to link on delete confirmation page

lory87:

- have to add this link as well on filer delete confirmation page

### DIFF
--- a/filer/templates/admin/filer/delete_selected_files_confirmation.html
+++ b/filer/templates/admin/filer/delete_selected_files_confirmation.html
@@ -43,6 +43,7 @@
                 {% endif %}
                 <input type="hidden" name="action" value="delete_files_or_folders">
                 <input type="hidden" name="post" value="yes">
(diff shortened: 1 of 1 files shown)


filer/templates/admin/filer/delete_selected_files_confirmation.html
```diff
@@ -43,6 +43,7 @@
                 {% endif %}
                 <input type="hidden" name="action" value="delete_files_or_folders">
                 <input type="hidden" name="post" value="yes">
+                <a href="#" onclick="window.history.back(); return false;" class="button cancel-link">{% trans "No, take me back" %}</a>
```
dhorelik: maybe `class='js-back-button'` and move the js out of here? I guess there can be other back buttons coming

lory87: it is implemented as on django and adminstyled as well. It will not be reused somewhere.

dhorelik: ok then

dhorelik:
lgtm

PR #793, opened 2016-04-05
focal point placement

lory87:

- fixes focal point calculation and position


filer/private/sass/components/_image-info.scss
```diff
@@ -171,11 +171,10 @@
...
         position: relative;
         z-index: 1;
-        width: 26px;
-        height: 26px;
+        width: 30px;
+        height: 30px;
         // makes sure that circle is positioned correctly and centered #684
-        margin-top: -15px;
-        margin-left: -15px;
+        margin: -15px;
```
vxsx: do you really need all of them set to -15? usually it doesn't really work as expected

lory87: yes focal point has to work as on the image above.

vxsx:
code LGTM

PR #802, opened 2016-04-08
fixes focal point position and if is NaN

lory87:

- I have to add overflow hidden to avoid position of focal point out of image if value is incorrect
- position focal point correctly over css
- fix value if is NaN


filer/static/filer/js/addons/focal-point.js
```diff
@@ -72,16 +65,20 @@ var Cl = window.Cl || {};
...
             }
 
+            if (isNaN(x) || isNaN(y)) {
+                return;
+            }
+
             this.circle.css({
-                'top': y - (circleSelectorSize / 2),
-                'left': x - (circleSelectorSize / 2)
+                'top': y,
```
vxsx: no need to quote keys
does the project have linting setup?

Brankub:
:+1:

vxsx:
:+1:

PR #809, opened 2016-04-12
not closing search on open dropdown

lory87:

- search should stay full width, when clicking on the arrow

rdluu939vv


filer/static/filer/js/base.js
```diff
@@ -48,7 +48,9 @@ Cl.mediator = new Mediator();
             if (event.type === 'focus') {
                 container.addClass('is-focused');
             } else {
-                container.removeClass('is-focused');
+                if (event.relatedTarget == undefined) {
```
vxsx: never `==` compare to undefined
should be something like `if (event.relatedTarget && event.relatedTarget.is(dropdownTrigger) {`

vxsx:
👍

PR #816, opened 2016-04-18
adds normal btn styles from mixin to action links

lory87:

-adds normal btn styles from mixin to action links with default buttons styles for active, hover, focus..

|BEFORE|AFTER|
|---|---|
|screen shot 2016-04-18 at 09 37 40|screen shot 2016-04-18 at 09 37 37|


vxsx:
👍

PR #820, opened 2016-04-18
upscale preview image to sidebar image width

lory87:

- fixes focal point if image is smaller thank preview size
- upscales to preview size


stefanfoulis:
👍

leonsmith:
This seem to have the side effect of huge memory consumption (multiple GBs to the point the machine OOM kills the process) & the preview page timing out.

Trying to debug the issue for the root cause but setting the height param to be 0 instead seems to resolve it whilst keeping the intent of this original fix

```python
        'admin_sidebar_preview': {
            'size': (SIDEBAR_IMAGE_WIDTH, 0), 
            'upscale': True
        },
```

PR #830, opened 2016-04-22
renames unfiled files with Unsorted Uploads

lory87:

- renames unfiled files folder name with Unsorted Uploads

### DIFF
--- a/filer/models/virtualitems.py
+++ b/filer/models/virtualitems.py
@@ -50,6 +50,7 @@ class UnfiledImages(DummyFolder):
     name = _("Unsorted Uploads")
     is_root = True
     _icon = "unfiled_folder"
+    is_unsorted_uploads = True
 
     def _files(self):
         return File.objects.filter(folder__isnull=True)
--- a/filer/models/virtualitems.py
+++ b/filer/models/virtualitems.py
@@ -47,7 +47,7 @@ class DummyFolder(mixins.IconsMixin):
 
 
 class UnfiledImages(DummyFolder):
-    name = _("unfiled files")
+    name = _("Unsorted Uploads")
     is_root = True
     _icon = "unfiled_folder"
 
--- a/filer/models/virtualitems.py
+++ b/filer/models/virtualitems.py
@@ -49,8 +49,8 @@ class DummyFolder(mixins.IconsMixin):
 class UnfiledImages(DummyFolder):
     name = _("Unsorted Uploads")
     is_root = True
-    _icon = "unfiled_folder"
     is_unsorted_uploads = True
+    _icon = "unfiled_folder"
 
     def _files(self):
         return File.objects.filter(folder__isnull=True)
--- a/filer/models/virtualitems.py
+++ b/filer/models/virtualitems.py
@@ -46,7 +46,7 @@ class DummyFolder(mixins.IconsMixin):
         return []
 
 
-class UnfiledImages(DummyFolder):
+class UnsortedImages(DummyFolder):
     name = _("Unsorted Uploads")
     is_root = True
     is_unsorted_uploads = True
@@ -83,7 +83,7 @@ class FolderRoot(DummyFolder):
 
     @property
     def virtual_folders(self):
-        return [UnfiledImages()]
+        return [UnsortedImages()]
 
     @property
     def children(self):
--- a/filer/templates/admin/filer/breadcrumbs.html
+++ b/filer/templates/admin/filer/breadcrumbs.html
@@ -11,11 +11,7 @@
             {% if ancestor_folder.label %}
                 {{ ancestor_folder.label }}
             {% else %}
-                {% if not ancestor_folder.is_unsorted_uploads %}
-                    {{ ancestor_folder.name }}
-                {% else %}
-                    {% trans "Unsorted Uploads" %}
-                {% endif %}
+                {{ ancestor_folder.name }}
             {% endif %}
         </a>
     {% endfor %}
--- a/filer/templates/admin/filer/breadcrumbs.html
+++ b/filer/templates/admin/filer/breadcrumbs.html
@@ -11,7 +11,7 @@
             {% if ancestor_folder.label %}
                 {{ ancestor_folder.label }}
             {% else %}
-                {% if ancestor_folder.name != "Unsorted Uploads" %}
+                {% if ancestor_folder.name != _("Unsorted Uploads") %}
                     {{ ancestor_folder.name }}
                 {% else %}
                     {% trans "Unsorted Uploads" %}
--- a/filer/templates/admin/filer/breadcrumbs.html
+++ b/filer/templates/admin/filer/breadcrumbs.html
@@ -11,7 +11,7 @@
             {% if ancestor_folder.label %}
                 {{ ancestor_folder.label }}
             {% else %}
-                {% if ancestor_folder.name != _("Unsorted Uploads") %}
+                {% if not folder.is_unsorted_uploads %}
                     {{ ancestor_folder.name }}
                 {% else %}
                     {% trans "Unsorted Uploads" %}
--- a/filer/templates/admin/filer/breadcrumbs.html
+++ b/filer/templates/admin/filer/breadcrumbs.html
@@ -11,7 +11,7 @@
             {% if ancestor_folder.label %}
                 {{ ancestor_folder.label }}
             {% else %}
-                {% if ancestor_folder.name != "unfiled files" %}
+                {% if ancestor_folder.name != "Unsorted Uploads" %}
                     {{ ancestor_folder.name }}
                 {% else %}
                     {% trans "Unsorted Uploads" %}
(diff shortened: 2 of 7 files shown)


vxsx:
seems good, but would appreciate backend review

lory87:
@vxsx yes I asked @stefanfoulis for a review

filer/templates/admin/filer/folder/directory_listing.html
```diff
@@ -130,7 +130,7 @@
                             <div class="navigator-breadcrumbs-folder-name-wrapper">
                                 <span class="navigator-breadcrumbs-folder-name">
                                     <span class="navigator-breadcrumbs-folder-name-inner">
-                                        {% if folder.name != "unfiled files" %}
+                                        {% if folder.name != "Unsorted Uploads" %}
```
sergii-kulakovskyi: if you compare this and `_("Unsorted Uploads")`, it could return `false` if it's not English

filer/templates/admin/filer/breadcrumbs.html
```diff
@@ -11,7 +11,7 @@
             {% if ancestor_folder.label %}
                 {{ ancestor_folder.label }}
             {% else %}
-                {% if ancestor_folder.name != "unfiled files" %}
+                {% if ancestor_folder.name != _("Unsorted Uploads") %}
```
cyberstar: this is an incorrect template syntax. there is no ``_`` function in the templates.
aslo I'm not sure that this will work with translated string. What will happen if current language is German, and the string ``Unsorted Uploads`` is translated (has a german translation instead of falling back to English?

filer/templates/admin/filer/folder/directory_listing.html
```diff
@@ -130,7 +130,7 @@
                             <div class="navigator-breadcrumbs-folder-name-wrapper">
                                 <span class="navigator-breadcrumbs-folder-name">
                                     <span class="navigator-breadcrumbs-folder-name-inner">
-                                        {% if folder.name != "unfiled files" %}
+                                        {% if folder.name != _("Unsorted Uploads") %}
```
cyberstar: same as above

filer/templates/admin/filer/folder/directory_listing.html
```diff
@@ -227,15 +227,15 @@
                                        data-max-uploader-connections="{{ uploader_connections }}">
                                         {% trans "Upload Files" %}
                                     </a>
-                                {% elif folder.name == "unfiled files" %}
+                                {% elif folder.name == _("Unsorted Uploads") %}
```
cyberstar: same as above

filer/templates/admin/filer/folder/directory_listing.html
```diff
@@ -227,15 +227,15 @@
...
                                     <a href="#" id="id_upload_button" title="{% trans 'Upload Files' %}"
                                        class="navigator-button navigator-button-upload js-upload-button"
                                        data-url="{% url 'admin:filer-ajax_upload' %}"
                                        data-max-uploader-connections="{{ uploader_connections }}">
                                         {% trans "Upload Files" %}
                                     </a>
                                 {% endif %}
-                                {% if folder.is_root and folder.name != "unfiled files" %}
+                                {% if folder.is_root and folder.name != _("Unsorted Uploads") %}
```
cyberstar: same as above

filer/templates/admin/filer/folder/directory_table.html
```diff
@@ -81,7 +81,7 @@
                             <div>
                                 <a href="{{ item.get_admin_directory_listing_url_path }}{% filer_admin_context_url_params %}"
                                     title="{% blocktrans with item.name as item_label %}Change '{{ item_label }}' folder details{% endblocktrans %}">
-                                    {% if item.name != "unfiled files" %}
+                                    {% if item.name != _("Unsorted Uploads") %}
```
cyberstar: same as above

filer/templates/admin/filer/breadcrumbs.html
sergii-kulakovskyi: @cyberstar why we can't use `_()` in templates? https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#string-literals-passed-to-tags-and-filters

stefanfoulis: So why don't we just make ``ancestor_folder.label`` or ``ancestor_folder.name`` return the correct translated string? Then all this logic would not be necessary.

lory87:
@cyberstar There are some more translations which we have to do. I will check them when they are available

stefanfoulis:
I'm not sure about capitalisation of ``"Unsorted Uploads"``. Shouldn't it be ``"unsorted uploads"``?

And we'll need to add translations for this new string at least for German, French and Italian.

lory87:
@stefanfoulis I will add and check translations there are some more to add. I added same capitalisation to have same value everywhere

stefanfoulis:
If it really is necessary to have special logic based on which virtual folder is being used, an explicit attribute could be added on the class here: https://github.com/divio/django-filer/blob/bugfix/rename-unfiled-files/filer/models/virtualitems.py#L49
Would be nicer than comparing strings.

cyberstar:
@lory87 @sergii-kulakovskyi  Ok, sorry, seems that i was wrong re ``_()`` in templates, i've deleted those.

filer/templates/admin/filer/folder/directory_listing.html
```diff
@@ -130,7 +130,7 @@
                             <div class="navigator-breadcrumbs-folder-name-wrapper">
                                 <span class="navigator-breadcrumbs-folder-name">
                                     <span class="navigator-breadcrumbs-folder-name-inner">
-                                        {% if folder.name != _("Unsorted Uploads") %}
+                                        {% if not folder.is_unsorted_uploads %}
                                             {{ folder.name }}
                                         {% else %}
                                             {% trans "Unsorted Uploads" %}
```
stefanfoulis: why not just ``{{ folder.name }}``? You can change that to be ``_("Unsorted Uploads")``. I don't think the ``if`` is needed.

stefanfoulis: actually, no need to even change it on the virtual Item. It already has the correct ``name`` attribute: https://github.com/divio/django-filer/blob/bugfix/rename-unfiled-files/filer/models/virtualitems.py#L50

The whole ``{% if .... %}...{% else %}...{% endif %}`` can be replaced by ``{{ folder.name }}``.

filer/templates/admin/filer/folder/directory_table.html
```diff
@@ -81,7 +81,7 @@
                             <div>
                                 <a href="{{ item.get_admin_directory_listing_url_path }}{% filer_admin_context_url_params %}"
                                     title="{% blocktrans with item.name as item_label %}Change '{{ item_label }}' folder details{% endblocktrans %}">
-                                    {% if item.name != _("Unsorted Uploads") %}
+                                    {% if not folder.is_unsorted_uploads %}
                                         {{ item.name }}
                                     {% else %}
                                         {% trans "Unsorted Uploads" %}
```
stefanfoulis: This ``{% if ..`` can also be replaced by a simple ``{{ item.name }}``.

filer/models/virtualitems.py
```diff
@@ -47,8 +47,9 @@ def logical_path(self):
 
 
 class UnfiledImages(DummyFolder):
```
stefanfoulis: We should change the class name to ``UnsortedUploads`` as well (and everywhere it is used in python code).

filer/templates/admin/filer/folder/directory_table.html
```diff
@@ -156,7 +152,7 @@
         </tbody>
     </table>
 
-    <div class="filer-dropzone-info-message js-filer-dropzone js-filer-dropzone-info-message hidden" data-url="{% if folder.id %}{% url 'admin:filer-ajax_upload' folder_id=folder.id %}{% else %}{% url 'admin:filer-ajax_upload' %}{% endif %}" data-folder-name="{% if folder.name and not folder.name == 'root' %}{{ folder.name }}{% else %}unfiled files{% endif %}" data-max-uploader-connections="{{ uploader_connections }}" data-max-file-size="20">
+    <div class="filer-dropzone-info-message js-filer-dropzone js-filer-dropzone-info-message hidden" data-url="{% if folder.id %}{% url 'admin:filer-ajax_upload' folder_id=folder.id %}{% else %}{% url 'admin:filer-ajax_upload' %}{% endif %}" data-folder-name="{% if folder.name and not folder.name == 'root' %}{{ folder.name }}{% else %}Unsorted Uploads{% endif %}" data-max-uploader-connections="{{ uploader_connections }}" data-max-file-size="20">
```
stefanfoulis: ``{% if folder.name and not folder.name == 'root' %}{{ folder.name }}{% else %}Unsorted Uploads{% endif %}``
Is this is the label of the dropzone when someone drags a file directly onto the "Folder List" view, right?

hmm... so if the folder has a name and the name is ``'root'``, then display ``Unsorted Uploads``?

Basically, whenever the current folder is not a real folder, it should say ``Unsorted Uploads``, right?

I'd say this can be simplified as well. @lory87 let's discuss in a hangout :-)

stefanfoulis:
👍 merge

PR #834, opened 2016-04-25
rename super lightest variable

lory87:

- updates and rename colors according to djangoCMS palette


vxsx:
👍

PR #839, opened 2016-05-03
Take hash into account when reloading

marcari:
On reload the query parameters were added after the hash. This PR fixes this bug by cutting the hash and putting it back after computing the new uri.


czpython:
@FinalAngel Can you have a look?

FinalAngel:
Thank you @marcari and sorry for the late reply. LGTM. Can you also adapt the Changelog?

PR #844, opened 2016-05-12
Change on_delete behaviour

glomium:
Djangos default behaviour is to cascade the deletion of a user, which results in the deletion of files and folders.

Also on_delete will become a required argument in Django 2.0.


czpython:
@glomium 
Hey there, thanks for the contribution.
Please update the migration files for both django and south. /cc @mkoistinen

PR #721, opened 2016-01-28
Remove all uses of `patterns`

mattbriancon:
Fixes #717

PR #850, opened 2016-05-23
Removed all usages of `django.conf.urls.patterns`

jsma:
This is a clean PR which can replace #721.


czpython:
@jsma Thanks for your contribution. As https://github.com/divio/django-filer/pull/721 is now completed, I'll close this as duplicate given that the former was first.

Issue #837, opened 2016-04-29
Errant migration added from byte literal


jschneier:
Hi. Just installed 1.2 using Python 3.5 and if I run a blank `makemigrations` without specifying a package a new filer migration is added. This is due to the byte literal added in [migration 0004](https://github.com/divio/django-filer/blob/develop/filer/migrations/0004_auto_20160328_1434.py). My recommendation would be to simply remove the byte literal.

yakky:
This has been fixed by 1.2.1

PR #852, opened 2016-05-23
Check FILER_IMAGE_MODEL setting before altering Image model

### LINKED ISSUE #851: KeyError in migration 0004 with custom/overridden FILER_IMAGE_MODEL
Migration 0004_auto tries to alter a field on the Image model. As this model can be overridden with the FILER_IMAGE_MODEL setting it may not exist and produce a KeyError: ('filer', 'image') when the migration is run.

Proposal: check in the migration if FILER_IMAGE_MODEL is set and only do the operation if it is not. The same is done in migration 0001 with CreateModel for the Image model.

aleehedl:
Fixes #851

### DIFF
--- a/filer/migrations/0004_auto_20160328_1434.py
+++ b/filer/migrations/0004_auto_20160328_1434.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from django.db import migrations, models
+from filer.settings import FILER_IMAGE_MODEL
 
 
 class Migration(migrations.Migration):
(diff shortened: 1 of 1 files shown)


filer/migrations/0004_auto_20160328_1434.py
```diff
@@ -11,10 +12,12 @@ class Migration(migrations.Migration):
...
-            field=models.CharField(blank=True, default=b'', max_length=64, verbose_name='subject location'),
-        ),
-    ]
+    operations = []
+    if not FILER_IMAGE_MODEL:
+        operations.append(
+            migrations.AlterField(
+                model_name='image',
+                name='subject_location',
+                field=models.CharField(blank=True, default=b'', max_length=64, verbose_name='subject location'),
```
mattbriancon: If you remove the byte string you can fix #837 too.

aleehedl: Done, thanks.

PR #854, opened 2016-06-01
Add missing migration

intellectronica:
This migration was left out of the latest version. It is being ignored by Django 1.8/1.9 and cms-helper (which is probably how it happened), but it brings the DB to the correct state (change of a default value to subject_location) and tests work in both cases.


FinalAngel:
👍

PR #856, opened 2016-06-02
Check FILER_IMAGE_MODEL setting before altering Image model in migrations

### LINKED ISSUE #855: KeyError in migration 0004 and 0005 with custom/overridden FILER_IMAGE_MODEL
Migrations 0004_auto and 0005_auto alter a field on the Image model. As this model can be overridden with the FILER_IMAGE_MODEL setting it may not exist and produce a KeyError: ('filer', 'image') when the migration is run.

Proposal: check in the migration if FILER_IMAGE_MODEL is set and only do the operation if it is not. The same is done in migration 0001 with CreateModel for the Image model.

aleehedl:
Fixes #855

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,6 +7,7 @@ next (20xx-xx-xx)
 * Rename filer picker widget upload button
 * Adds missing @2x icon files
 * Added missing migration #854
+* Adds a check for FILER_IMAGE_MODEL setting before altering image model in migrations
 
 1.2.0 (2016-04-26)
 ------------------
--- a/filer/migrations/0004_auto_20160328_1434.py
+++ b/filer/migrations/0004_auto_20160328_1434.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from django.db import migrations, models
+from filer.settings import FILER_IMAGE_MODEL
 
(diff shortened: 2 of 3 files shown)


FinalAngel:
thank you @aleehedl would you mind also adding a change log entry in: https://github.com/divio/django-filer/blob/develop/CHANGELOG.txt and we're good to go.

aleehedl:
@FinalAngel, changelog entry added.

skirsdeda:
@FinalAngel FILER_IMAGE_MODEL setting as it is poses some more problems with migrations. This is because it is practically a swappable model (just like User model in django.contrib.auth) but filer treats it differently . I am currently working on a branch which uses FILER_IMAGE_MODEL setting for Image.META.swappable attribute and therefore makes Image model properly swappable. It solves this issue and many others by better design instead of a bunch of ifs.

So I would recommend refraining from merging this and instead looking at a proper solution.

skirsdeda:
@aleehedl, @FinalAngel Check out #857.

mkoistinen:
This will likely be superseded by #857.

stefanfoulis:
Closing since it is now fixed by #857 . Please re-open if it is still an issue.

PR #869, opened 2016-06-23
Admin: directory-listing permissions speedup

### LINKED ISSUE #868: Slow directory listing in admin for folder with lots of files
I've run into a performance issue with the admin page when viewing a folder that contain lots of files (hundreds, but not thousands). It's fine if the current user is a super user, but for a non super it slows to a crawl, taking upwards of 30 seconds to load a page.

I think I tracked this down to be caused by [this line](https://github.com/divio/django-filer/blob/1.2.0/filer/admin/folderadmin.py#L366), and I think the reason is that we are doing a permissions check for every item in the list, before we give it to the paginator. But all we actually want is to know if we should show the "Change" link in `directory_table.html`, based on `item_perms.change`, so we only need to do this check for each item we are showing on the page, which is just 20. The reason it is fast for a superuser is that they skip any permissions checking that hits the database.

This change could be done either in Python, but after the paginator has chopped up the list, or in the template. I'll have a play about with it and see where I get.

plumdog:
Fixes #868.

Instead of checking to see if we should show the permissions link for all items, only do that check for items that we are displaying.

Adds a template tag that tells us whether we have permission to do an action on the item. This could be moved into Python if we rather, but would mean faffing about with the items inside the page that the paginator gives us, which seemed less Django-ish.

### DIFF
--- a/filer/templatetags/filer_admin_tags.py
+++ b/filer/templatetags/filer_admin_tags.py
@@ -34,3 +34,20 @@ def filer_admin_context_hidden_formfields(context):
         '<input type="hidden" name="{0}" value="{1}">',
         admin_url_params(request).items(),
     )
+
+
+@register.assignment_tag(takes_context=True)
+def filer_has_permission(context, item, action):
+    """Does the current user (taken from the request in the context) have
+    permission to do the given action on the given item.
+
+    """
+    permission_method_name = 'has_{action}_permission'.format(action=action)
+    permission_method = getattr(item, permission_method_name, None)
+    request = context.get('request')
+
+    if not permission_method or not request:
+        return False
+    # Call the permission method.
+    # This amounts to calling `item.has_X_permission(request)`
+    return permission_method(request)
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -360,8 +360,7 @@ class FolderAdmin(PrimitivePermissionAwareModelAdmin):
             folder_files.sort()
 
         items = folder_children + folder_files
-        items_permissions = [(item, {'change': self.has_change_permission(request, item)}) for item in items]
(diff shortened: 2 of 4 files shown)


plumdog:
Also, Travis appears to have done nothing, unsuccessfully.

yakky:
LGTM
@czpython @mkoistinen how about adding this to 1.2.3 release?

yakky:
@plumdog would you rebase on current develop? This should fix the test failure

plumdog:
@yakky rebased with develop. Have also had to tweak the tests to fix.

yakky:
LGTM
@czpython I'd like to merge this if it looks good for you

filer/templatetags/filer_admin_tags.py
```diff
@@ -34,3 +34,23 @@ def filer_admin_context_hidden_formfields(context):
...
+    """
+    permission_method_name = 'has_{action}_permission'.format(action=action)
+    permission_method = getattr(item, permission_method_name, None)
+    request = context.get('request')
+
+    if not permission_method or not request:
+        return False
+    # Try to call the permission method.
+    # This amounts to calling `item.has_X_permission(request)`
+    try:
```
czpython: the original implementation does not do this, please avoid such wide error handling.
If this method fails then it should fail loudly.

czpython:
Thanks for this.
Overall lgtm, it can land on the next release once the requested change is addressed.

yakky:
@plumdog would you be able to address paulo's comment?

filer/templatetags/filer_admin_tags.py
plumdog: Agreed and fixed.

plumdog:
@yakky Changes made.

yakky:
@plumdog :+1: merging
Thanks for contributing this!

PR #871, opened 2016-06-23
Fixed `file_ptr` issues in Django 1.9+

mkoistinen:


stefanfoulis:
👍 
(but I thought the tests passed fine without the overridden file_ptr field. Why is it necessary?)

filer/models/abstract.py
```diff
@@ -38,6 +42,8 @@ class BaseImage(File):
 
     subject_location = models.CharField(_('subject location'), max_length=64, blank=True,
                                         default='')
+    if DJANGO_GTE_19:
+        file_ptr = models.OneToOneField(to='filer.File', related_name='+')
```
yakky: @mkoistinen any specific reason for not using ``related_name='%(app_label)s_%(class)s_file'`` here?

PR #876, opened 2016-06-27
size, sha1 and width, height field recalculation overhaul

skirsdeda:
This is a complete rework of part of #500 relating to size, sha1, width and height recalculation. Current implementation is based on how Django's ImageFileField recalculates width, height. Basically, instead of updating these fields in `save()`, they are updated immediately after MultiStorageFileField value is set.

I also improved change detection in descriptor a bit by checking whether field value is actually changed and detecting calls originating from model's `__init__` properly.

### DIFF
--- a/filer/fields/multistorage_file.py
+++ b/filer/fields/multistorage_file.py
@@ -7,6 +7,7 @@ import warnings
 from io import BytesIO
 
 from django.core.files.base import ContentFile
+from django.db.models.fields.files import FileDescriptor
 from django.utils import six
 from easy_thumbnails import fields as easy_thumbnails_fields
 from easy_thumbnails import files as easy_thumbnails_files
@@ -38,6 +39,32 @@ def generate_filename_multistorage(instance, filename):
         return upload_to(instance, filename)
     else:
         return upload_to
+
+
+class MultiStorageFileDescriptor(FileDescriptor):
+    """
+    This is rather similar to Django's ImageFileDescriptor.
+    It calls <field name>_data_changed on model instance when new
+    value is set. The callback is suposed to update fields which
+    are related to file data (like size, checksum, etc.).
+    When this is called from model __init__ (prev_assigned=False),
+    it does nothing because related fields might not have values yet.
+    In such case data_changed callback should be called at the end of model __init__
+    (File.__init__ in this case).
+    """
+    def __set__(self, instance, value):
+        prev_assigned = self.field.name in instance.__dict__
+        previous_file = instance.__dict__.get(self.field.name)
+        super(MultiStorageFileDescriptor, self).__set__(instance, value)
+
+        # To prevent recalculating file data related attributes when we are instantiating
+        # an object from the database, update only if the field had a value before this assignment.
+        # To prevent recalculating upon reassignment of the same file, update only if value is
+        # different than the previous one.
+        if prev_assigned and value != previous_file:
+            callback_attr = '%s_data_changed' % self.field.name
+            if hasattr(instance, callback_attr):
+                getattr(instance, callback_attr)()
 
 
 class MultiStorageFieldFile(ThumbnailerNameMixin,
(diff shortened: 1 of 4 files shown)


skirsdeda:
@mkoistinen This is good to be merged. I tested this pretty hard and it is now more sophisticated than Django's ImageField dimension recalculation logic.

filer/fields/multistorage_file.py
```diff
@@ -40,6 +41,32 @@ def generate_filename_multistorage(instance, filename):
...
+        previous_file = instance.__dict__.get(self.field.name)
+        super(MultiStorageFileDescriptor, self).__set__(instance, value)
+
+        # To prevent recalculating file data related attributes when we are instantiating
+        # an object from the database, update only if the field had a value before this assignment.
+        # To prevent recalculating upon reassignment of the same file, update only if value is
+        # different than the previous one.
+        if prev_assigned and value != previous_file:
+            callback_attr = '{}_data_changed'.format(self.field.name)
+            if hasattr(instance, callback_attr):
```
yakky: I might be worried that this could be a performance hit (hastattr/getattr are generally slow). Maybe catching the exception? If I understand correctly this is only called when updating the underlying image, so performance might not be an issue (the IO past this point will be order of magnitude slower)

filer/models/filemodels.py
```diff
@@ -81,6 +81,35 @@ def matches_file_type(cls, iname, ifile, request):
...
+        self.file_data_changed(post_init=True)
+
+    def file_data_changed(self, post_init=False):
+        """
+        This is called whenever self.file changes (including initial set in __init__).
+        MultiStorageFileField has a custom descriptor which calls this function when
+        field value is changed.
+        Returns True if data related attributes were updated, False otherwise.
+        """
+        if hasattr(self, '_file_data_changed_hint'):
```
yakky: Why not using ``True``/``False``/``None`` instead of adding and removing the attribute?

yakky:
@skirsdeda sorry it took so much time to review this. Would you care to check the comments above?

filer/fields/multistorage_file.py
```diff
@@ -40,6 +41,32 @@ def generate_filename_multistorage(instance, filename):
...
+    It calls <field name>_data_changed on model instance when new
+    value is set. The callback is suposed to update fields which
+    are related to file data (like size, checksum, etc.).
+    When this is called from model __init__ (prev_assigned=False),
+    it does nothing because related fields might not have values yet.
+    In such case data_changed callback should be called at the end of model __init__
+    (File.__init__ in this case).
+    """
+    def __set__(self, instance, value):
+        prev_assigned = self.field.name in instance.__dict__
```
czpython: this is redundant as `previous_file` below can be used to know if there was a previous file assigned

filer/fields/multistorage_file.py
```diff
@@ -40,6 +41,32 @@ def generate_filename_multistorage(instance, filename):
...
+        previous_file = instance.__dict__.get(self.field.name)
+        super(MultiStorageFileDescriptor, self).__set__(instance, value)
+
+        # To prevent recalculating file data related attributes when we are instantiating
+        # an object from the database, update only if the field had a value before this assignment.
+        # To prevent recalculating upon reassignment of the same file, update only if value is
+        # different than the previous one.
+        if prev_assigned and value != previous_file:
```
czpython: change to `if previous_file and value != previous_file`

filer/models/filemodels.py
czpython: Agree, we should always have this hint attribute set and work with its value instead

skirsdeda:
@yakky I'll fix this soonish.

filer/fields/multistorage_file.py
skirsdeda: @czpython I'm not sure if this is redundant because ```previous_file``` would be ```None``` both if it is not yet assigned and if it is assigned ```None```.  ```prev_assigned``` on the other hand would be true if field was assigned ```None```.

filer/fields/multistorage_file.py
skirsdeda: @yakky This is only called when file data is actually changed (assigned another file). In case when called while instantiating from DB, the upper 'if' is false. So I would say performance hit is negligible.

skirsdeda:
@yakky changed ```_file_data_changed_hint``` to default to None. See my comments for other change requests.

yakky:
@skirsdeda thanks, that looks good

yakky:
@skirsdeda thanks for this!

PR #882, opened 2016-07-01
Prevent automatically creating new migration

vstoykov:


yakky:
@vstoykov thanks for this cotnribution

PR #895, opened 2016-07-24
Fixed canonical urls

### LINKED ISSUE #896: Canonical URL doesn't work under Windows
```
>>> from filer.models.filemodels import File
>>> f = File.objects.first()
>>> f.canonical_url
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "D:\Temp\mysite\env\lib\site-packages\filer\models\filemodels.py", line 2
65, in canonical_url
    'uploaded_at': self.uploaded_at.strftime('%s'),
ValueError: Invalid format string
```

TheDeadOne:
`%s` is not supported format string for `strftime`, it is not portable, it may silently produce a wrong result for an aware `datetime` object, it fails if an input is in UTC but local timezone is not UTC. The only reason it's working (on unix-like systems) is because Python is passing the information to system's `strftime`.

Fixes #896

### DIFF
--- a/filer/models/filemodels.py
+++ b/filer/models/filemodels.py
@@ -4,6 +4,8 @@ from __future__ import absolute_import, unicode_literals
 
 import hashlib
 import os
+import pytz
+from datetime import datetime
 
 from django.conf import settings
 from django.core import urlresolvers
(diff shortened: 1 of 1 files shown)


jschneier:
(Trying to get 1.2.6 released :p)

Change `pytz.utc` to `timezone.utc` and the import to be `from django.utils import timezone`.

yakky:
Issues fixed in #923

PR #909, opened 2016-08-30
Added aldryn config

czpython:


FinalAngel:
JSON +1, cannot say too much about the python code there cause me noob

aldryn_config.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def to_settings(self, data, settings):
+        from functools import partial
+        from aldryn_addons.utils import boolean_ish, djsenv
+        from aldryn_django import storage
+
+        env = partial(djsenv, settings=settings)
+
+        # django-filer
+        settings['FILER_DEBUG'] = boolean_ish(env('FILER_DEBUG', settings['DEBUG']))
```
czpython: Agreed. For this first stage I think it's fine as we've not provided these options before.

PR #914, opened 2016-09-22
Fixed markup issue with editing file in admin

vxsx:
before:
screen shot 2016-09-22 at 17 07 49

after:

screen shot 2016-09-22 at 17 07 43


czpython:
👍

PR #918, opened 2016-10-10
Upgrade django

jschneier:
Based on #911.

Also, the various compatibility helpers _litter_ the code base. Are we okay to drop Django < 1.8? Obviously in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,16 +23,19 @@ env:
   - TOX_ENV=py27-dj17-thumbs2x
   - TOX_ENV=py27-dj18-thumbs2x
   - TOX_ENV=py27-dj19-thumbs2x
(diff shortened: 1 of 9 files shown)


jschneier:
Okay this is @mattbriancon fixes with the tests now passing.

@czpython let me know if you need anything else. Also, like I mentioned in my PR I'd be happy to go through and remove the Django < 1.8 compatibility helpers. It's a lot of code that I think adds a decent maintenance burden.

mattbriancon:
@jschneier apologies for letting my PR slip and thanks for putting this together!

yakky:
@jschneier @mattbriancon thanks for contributing this. We can drop 1.7- in 1.3, which has no timeline yet, bue I agree we should cut old compatibility soon

PR #922, opened 2016-10-16
Add Django 1.10 migration

### LINKED ISSUE #920: Need to add migrations to go with Django 1.10 changes

yakky:
Fix #920


yakky:
@czpython are you ok with this hackish solution? This is generated automatically from Django 1.10 but it's not compatible with 1.9 and below. So just ignoring it's the only viable solution to me

czpython:
@yakky What happens if this runs on versions < 1.10? Altermodeloption has been around since 1.7

yakky:
Altermodeloption is not the issue, but rather the ``'default_manager_name'``  which has been introduced in Django 1.10.
In Django 1.9 and below the migration is simply ignored. Intstruction does not alter in any way the database schema, and thus it's safe to ignore on Django versions which does not support it

czpython:
@yakky ok to merge but please add a comment on the condition as to why we're doing this.

yakky:
@czpython done. Will merge at the end of the test run

jschneier:
Yep I was going to do the same thing.

jschneier:
Oh the one thing I'll note is that maybe we shouldn't have a hard dependency on:
 
`from filer.utils.compatibility import GTE_DJANGO_1_10`

We won't be able to remove that in the future then (say when Django 2.0) or whatever comes out and we don't support Django < 1.10.

yakky:
Why shouldn't be able to remove the import?

jschneier:
We could. I'm worried about an unlikely situation, maybe impossible (we remove the compat helper but forget to remove it from here and that breaks the app for everyone). I guess we'll want to remove the if check regardless so it's fine.

yakky:
tests runs with migration activated, thus if we remove the compat helper the tests will fail
In case we will remove 1.9 support, we can just remove the check and let migration run in any case

PR #923, opened 2016-10-16
Merge #895

yakky:
Fix #895 

Fixed a couple of issues with the original PR

PR #925, opened 2016-10-20
Fix new canonical when USE_TZ is False

yakky:
A fix for a regression introduced in #923


yakky:
@czpython please review

PR #926, opened 2016-10-20
Fix history link in Django 1.9+

yakky:
In Django 1.9 history link is at the same level of the changeview


yakky:
@czpython please review

czpython:
@yakky is there a way we can just reverse the url by name?

yakky:
@czpython yes, we still need to inject the history view name in the context, though

czpython:
That's ok for me 👍

yakky:
@czpython done, have a look at it

PR #941, opened 2016-11-24
fixes object tool placement on image detail page

lory87:
Before
screen shot 2016-11-24 at 14 42 37

After
screen shot 2016-11-24 at 15 29 15


filer/private/sass/components/_image-info.scss
```diff
@@ -34,6 +34,10 @@
         margin-bottom: 20px;
         padding: 25px;
         border-radius: 0;
+        // fixes object tools placement on image detail page
+        + #content-main .object-tools {
+            top: -310px;
```
FinalAngel: does that also work on popups and other places, that magic number seems a bit weird.

PR #945, opened 2016-11-29
Replaces django's extended unquote function with urlunquote

kirberich:
This avoids wrongly unescaping characters that aren't escaped by the browser, like underscores.


yakky:
@kirberich could you provide a test, or at least a way to reproduce the error and the fix?

yakky:
@kirberich super thanks!. Sorry it took me a while to get back to you

PR #948, opened 2016-12-29
Make localization for fieldsets of ImageAdmin

listeng:


yakky:
@listeng thanks for this!

Issue #954, opened 2017-01-13
Orphaned files if file is replaced via Advanced -> Replace in admin 


SachaMPS:
We noticed that we have lots of orphaned media files in filer_public/. 

The root cause of this seems to be if a user replaces a file via Admin > Adnvanced > File:
https://github.com/divio/django-filer/blob/develop/filer/admin/fileadmin.py#L55

Replacing a file won't delete the old file and won't delete all depending files like created thumbnails.

It seems that this is a very well known problem by django itself:
http://stackoverflow.com/questions/4394194/replacing-a-django-image-doesnt-delete-original
http://stackoverflow.com/questions/4787141/django-how-to-replace-overwrite-update-change-a-file-of-filefield
https://code.djangoproject.com/ticket/11663#no1

So my question is: Can this be regarded as a bug and should it be fixed in django-filer?

SachaMPS:
I've tried one of the suggestions from stackoverflow... Adding this to the save method of filer.models.File seems to do the trick: 

` 
    try:
        this = File.objects.get(id=self.id)
        if this.file != self.file:
            this.file.delete(save=False)
    except:
        pass`

As far as i understand it would make sense to add this behaviour. Either no files should be deleted at all or all files should be deleted on every action.

jrief:
Although, it doesn't fix this bug, you may manually apply this pending PR https://github.com/divio/django-filer/pull/912. It adds a management command to remove orphaned file.

SachaMPS:
@jrief Thank you... actually I already stumbled over your PR and right now modifying it to delete also the thumbnails which are located in filer_public_thumbnails.

jrief:
@SachaMPS if you have some time to work on this pull request, it would be great. I think @yakky 's proposal to add one command named ``filer_check`` with these options

``--orphans``: scan for orphans
``--missing``: scan for missing
``--delete-orphans``: delete orphans
``--verbose=3``: output verbose string instead of raw filenames

would fulfill all of our requirements.

yakky:
@SachaMPS while a bit  brutal, I think that your solution is better than keeping orphaned files around. Do you mind to open a PR for your change?
Also: any help on #912 is very much appreciated

merwok:
This is still relevant.

merwok:
reopen

PR #958, opened 2017-02-06
#954 avoid orphaned files when using admin > advanced > replace file

SachaMPS:
This should be a workaround for:
https://github.com/divio/django-filer/issues/954

### DIFF
--- a/filer/models/filemodels.py
+++ b/filer/models/filemodels.py
@@ -191,6 +191,13 @@ class File(PolymorphicModel, mixins.IconsMixin):
         if self._old_is_public != self.is_public and self.pk:
             self._move_file()
             self._old_is_public = self.is_public
+        # Delete old file(s) when updating the file via: admin > advanced > replace file
+        try:
+            this = File.objects.get(id=self.id)
+            if this.file != self.file:
+                this.file.delete(save=False)
+        except:
+            pass
         super(File, self).save(*args, **kwargs)
     save.alters_data = True
 


benzkji:
not a real workaround, looks ok?! slightly better exception handling could be done, but then that's it - I once read one should not just `except:`, but always provide what you expect to except...

+1 for merging this and wasting less server space ;-)

filer/models/filemodels.py
```diff
@@ -191,6 +191,13 @@ def save(self, *args, **kwargs):
         if self._old_is_public != self.is_public and self.pk:
             self._move_file()
             self._old_is_public = self.is_public
+        # Delete old file(s) when updating the file via: admin > advanced > replace file
+        try:
+            this = File.objects.get(id=self.id)
+            if this.file != self.file:
+                this.file.delete(save=False)
+        except:
```
yakky: It's better to explicitly handle the expected exceptions

yakky:
@SachaMPS Thanks for contributing this,, could you rebase on top of current develop?
Sorry it took so long to handle this

benzkji:
how about a merge...1.4 is out :-)

jrief:
@benzkji I fully agree with @yakky on https://github.com/divio/django-filer/pull/958#discussion_r178407709

benzkji:
sure!

benzkji:
is this issue resolved? anyone?

merwok:
No, this PR was not merged (author did not reply to review) and the ticket is still open.

benzkji:
someone would need to handle the exception, and off we go, right?

marksweb:
What are the possible exceptions here. Obviously there's the `DoesNotExist` from the `get()`, but is there something else?

benzkji:
if `this.file` is none (not the usual case, but could be...), this.file.delete() would throw a "None has no method 'delete()'". Rather edge case. Dont see others...

benzkji:
This is important, isn't it?

benzkji:
@jrief or was this alreaedy merge before? It's an easy picking...

PR #962, opened 2017-02-17
Remove wrong comment

fengsi:
This old comment is extremely misleading (especially when you have custom template)!  :(

See commit https://github.com/divio/django-filer/commit/3718694f73e4560f5ae8bded356987acd85b55a1


czpython:
Thanks!

PR #964, opened 2017-02-20
Related Names work on the opposite direction of FK

nidhog:
Example:
```
company.logo  # this is the logo of the company
logo.logo_company  # these are the companies with that logo
```


czpython:
Thanks!

PR #967, opened 2017-03-01
Updated translations

czpython:


FinalAngel:
@czpython please pull again from transifex, updated some translations

PR #970, opened 2017-03-07
Fix regression introduced in fc324e536f0d7097b9083c0dcf0c217d4c6d9d69

### LINKED ISSUE #965: Image _width and _height are NULL after import from fixture
How to reproduce:

On a site with images

```
./manage.py dumpdata filer.Image > images.json
./manage.py loaddata images.json
```

check the database, in table ``filer_image`` fields ``_width`` and ``_height`` are NULL after the import, whereas the fixture contains the correct sizes.

One workaround is to reset the sizes with this simple script

```
from filer.models.imagemodels import Image

for img in Image.objects.all():
    img.file_data_changed()
    img.save()
```

Used versions:
Django==1.10.5
django-filer==1.2.6

jrief:
This PR fixes #965 which was introduced by @skirsdeda in fc324e536f0d7097b9083c0dcf0c217d4c6d9d69

Are there any special reasons why you set ``_width`` and ``_height`` to ``None``?


skirsdeda:
@jrief well this is questionable. So if file has been changed and is inaccessible, would you keep `width` and `height` set to what was true for a previous file? I think what really has to be fixed is change detection when importing from a fixture.

jrief:
The method ``file_data_changed()`` is invoked for every image imported through ``./manage.py loaddata``. You should not modify this data. This is the reason why Django's signal handler introduced the ``raw`` attribute. The docs state:

> raw
> A boolean; True if the model is saved exactly as presented (i.e. when loading a fixture). One should not query/modify other records in the database as the database might not be in a consistent state yet.

your proposal to fix this is

> I think what really has to be fixed is change detection when importing from a fixture.

but where do we see if ``save()`` was invoked by by loaddata or through program code?

Anyway, what problem did it solve to set ``_width`` and ``_height`` to ``None``?

skirsdeda:
@jrief setting `_width` and `_height` to `None` solved the problem of having valid data. You can change the underlying file for filer's image and `_width`/`_height` should always match the actual file.

Regarding `raw` attribute, I think there must be a way to detect that in `save()`. After all, signals are invoked within `save()`.

jrief:
> You can change the underlying file for filer's image

How can I do this, while loading a fixture? Django doesn't provide any hooks for this.

skirsdeda:
You cannot do this while loading a fixture. But you can do this when editing filer's file via admin interface. Hence this fix should somehow detect if model is being loaded from a fixture and only then skip `file_data_changed`.

jrief:
Users expect that the data they dumped into a fixture is restored after loading it.

Or shall I tell them that after loading a fixture they have to go to the admin and reset the size for every image in that fixture? Sometimes there are hundreds!

Since *you* introduced that regression, it's *you* which should come up with a solution which works in a usable manner.

jrief:
@stefanfoulis @yakky shall this regression now become a feature?

jrief:
@skirsdeda 
Where do you need ``_width`` and ``_height`` set to ``None``, if the image was not found? There you should use another marker anyway.

skirsdeda:
@jrief 
I have looked into how loaddata works. It skips `save` and calls `save_base` instead (with `raw=True`). So the best strategy would be to postpone changing `_width`, `_height` (and size, sha1 for that matter) until `save` is called on the model. That way, when loaddata is used, `save` would never be called and these fields wouldn't be overwritten.

jrief:
Invoking ``loaddata`` does not call ``save_base``, hence this is not an option. You can try this yourself, by overriding the ``save_base`` method in the model – when performing ``loaddata`` it is never invoked!

I now postponed the optional reset of ``_width`` and ``_height`` for invalid images into the ``save`` method. This btw. is where is belongs to! This should be done before saving and not during the construction of the object, as it was up to now.

@yakky please review my updated pull request.

jrief:
This fix should work now.
If the image file is missing, then use that one provided by ``self.file_ptr.file``, which is already available while loading a fixture.

I also added the image part to the unit test, in order to check if the reimport of the fixture works.

jrief:
unfortunately this patch did not work anymore on Django-1.11, because there the related file object `self.file.file` is not ready yet.
I therefore added a check: During initialisation (via fixture), an existing `_width` and `_height` are not overridden anymore.
Otherwise we get a regression of #965, which means that it is impossible to load files and images dumped through a fixture.

jrief:
I just created a new pull request #1034 to fix this newly arising problem.

PR #981, opened 2017-05-23
Fix issue with inlines

blablacio:
When dynamically adding dropzones (i.e. through inlines) they don't get initialized. This is caused by fix in baa8d52b31cf16d2145a2a8d9328ed74692dd361. Tested against Django 1.11.1 only.

### DIFF
--- a/filer/static/filer/js/addons/dropzone.init.js
+++ b/filer/static/filer/js/addons/dropzone.init.js
@@ -147,8 +147,8 @@ django.jQuery(function ($) {
             Dropzone.autoDiscover = false;
         }
         dropzones.each(createDropzone);
-        $('.add-row a').on('click', function () {
-            var dropzones = $(dropzoneSelector);
+        $(document).on('formset:added', function (ev, row) {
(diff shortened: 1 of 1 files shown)


bashu:
@blablacio still not merged?

blablacio:
@bashu Doesn't seem to have been merged. You could just use my branch for the time being...

yakky:
@blablacio thanks for contributing this. Fix does not work in Django 1.8. As we are still compatible with it in 1.3, we should somehow find a fix that works in 1.8 and 1.9+ (or run one of the two versions of the code according to the django version)
Could you have a look at this?

yakky:
After a bit of digging I would say that `window.__admin_utc_offset__` (which is set by Django admin in 1.8 and 1.9+ i replaced by a data attribute) looks like a good candidate to detect the django version in the js code

yakky:
Fixed by #1060 
Thanks @blablacio for proposing the fix. I merged a solution compatible with all the supported django versions

Issue #982, opened 2017-05-26
Migration 0007_auto_20161016_1055 raises key error when custom Image model is provided


striketeamjezuz:
# Issue

Setting **FILER_IMAGE_MODEL** to a custom image model causes migration 0007_auto_20161016_1055 to raise a key error when attempting to run makemigrations on another app. This error may also be raised in other cases but running makemigrations is the only one I have explored. The migration tries to change the model options of filer.Image instead of the custom image model provided. The custom model I provided is identical to the default image model accept for an additional datetime field and the Meta.app_label set to the name of my own app. Maybe there is something I am missing.

Stack trace:
```
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/core/management/commands/makemigrations.py", line 149, in handle
    loader.project_state(),
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/db/migrations/loader.py", line 317, in project_state
    return self.graph.make_state(nodes=nodes, at_end=at_end, real_apps=list(self.unmigrated_apps))
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/db/migrations/graph.py", line 409, in make_state
    project_state = self.nodes[node].mutate_state(project_state, preserve=False)
  File "/home/benjamin/projects/virtualenvs/deathstar/lib/python3.4/site-packages/django/db/migrations/migration.py", line 92, in mutate_state
    operation.state_forwards(self.app_label, new_state)
[...]

PR #857, opened 2016-06-03
Make Image model swappable using a Meta.swappable option

skirsdeda:
Custom Image models are still problematic mostly because of issues with migrations. This is because Image model swappability implementation in filer is sort of broken by design :) I propose switching to a standard Django way for model swapping - Meta.swappable option.

The branch in it's current form is already working but I'm marking is as a Work in progress because there are some problems:
- FILER_IMAGE_MODEL setting now uses django content type instead of full python path to model. This breaks compatibility, however it's very easy to fix.
- FILER_FILE_MODELS setting default is currently broken for the same reason. I think it would be a good idea to switch FILER_FILE_MODELS setting to content type syntax as well but this is not for me to decide.

PR #989, opened 2017-06-26
Guard migration for custom filer image model

jdzero:
https://github.com/divio/django-filer/issues/982


jdzero:
May also be worth writing a test that installs a CustomImage to ensure future migrations are exercised.  If this is of interest I will add it to the PR.  Thanks!

skirsdeda:
All the problems regarding custom image model have been solved comprehensively in #856. And it has been sitting for a looong time now.

stefanfoulis:
Closing since it is now fixed by #857

PR #993, opened 2017-07-13
Draft/Live for Files

stefanfoulis:
# Publisher (draft/live)

- [x] vendor in latest changes from djangocms-publisher
  - [ ] fix bugs in folder list view actions
- [ ] update to cms pagetree indication style in folder list view (wait for implementation in aldryn-newsblog/djangocms-publisher to mature a bit first)

## general
- [x] rename "draft" to "not published"  @stefanfoulis
- [x] rename "pending changes" to "unpublished changes"  @stefanfoulis
- [x] default string representation of objects should contain "[not published]" at the end.  @stefanfoulis
## list view (Directory Browser)
- [ ] list view action: publish @stefanfoulis
- [x] "pending deletion" in default red @vxsx
- [x] label in same warning yellow as in cms @vxsx
## detail view
- [x] "pending deletion" label on detail view as well 
- [x] remove "cancel" on draft detail @stefanfoulis 
- [x] "request deletion" at the left with bucket icon
- [x] "discard deletion" at the left
- [x] "edit" blue right
- [x] "cancel" on live on the right (to the left of "edit")
- [x] "publish"/"publish deletion": tooltip "Permissions needed"
## picker
- ~~drafts should have a disabled arrow picker with "not published yet" tooltip at the front~~
- ~~find another icon for the pick. file label no longer picks~~ (to be consistent with folder)
  we'll remove the icon
- [ ] clicking on "not published yet" or "unpublished changes" should allow navigating and publishing without loosing the "picker" context.
- [x] pick draft version from detail page. right of file name. "Choose file" button
- [x] pick folder on top of folder list.
- [x] "choose this file" as a button next to file name on file detail. (don't loose the picker context when publishing/navigating!)
    - [x] fix missing condition to show this button only in the picker context @stefanfoulis 
- [ ] submit-row actions and navigating/saving files should not loose picker context @stefanfoulis

## other
- [ ] tests
- [ ] docs
- [ ] styling/html that looks good with and without djangocms-admin-style
- [ ] make ``FILER_NEW_FILE_DRAFTS_STATUS`` not influence migrations

## Phase 2 / later / maybe
- [ ] admin history does not contain "publish" actions. It should.
- [ ] update with latest version from divio/djangocms-publisher#2 when it is done (vendor in again?)

## Done

- [x] Only show indication for drafts, not for live objects (less clutter. Most object will be live most of the time anyway). This also makes the mode where draft/live is deactivated in filer easier.
- [x] integrate DraftMixin in File model
- [x] basic admin integration
- [x] rewrite linking objects from draft to live (on publish and discard draft). For the case where cms plugins or other things are linking to the draft version which is about to be deleted. Prevents ``SET_NULL`` and ``CASCADES`` on those objects.
- [x] setting to disable draft/live functionality globally
- [x] setting to configure if new uploads are draft or live
- [x] a global "publish filer file" permission with check
- **admin detail view**
   - [x] custom submit-row with the draft/live related actions. hide save/delete sometimes.
   - [x] enforce permission denied for save/delete on read-only objects (live versions)
   - [x] readonly for live obj
   - [x] show delete view when deleting a draft that has not been published yet
   - [x] show delete view before the "publishdeletion" action
   - [x] don't load subject location widget in readonly (live) view
   - [x] better markup for submit-row buttons that are actually just links ("edit draft", "show live")
   - [x] show draft indicator
   - [x] better button labels
   - [x] translateable button labels
   - [x] Show cross links (from draft to live and live to draft) with the buttons. Maybe as a subtle link in the breadcrumbs?
- **admin directory listing**
  - [x] show drafts with public version nicer
  - [x] *delete* link should "request deletion" instea of delete for live objects
  - [x] update "copy" operation (needs to set ``live=None`` for drafts)
  - [x] show pending changes indicator on live files
  - [x] show pending deletion indicator
- [x] Don't show both the live and draft file side-by-side in the directory listing. Instead only show one of the versions and show the draft as an indicator and part of the live file.
- [x] Divio Cloud integration (``aldryn_config.py``)


czpython:
Replaced by https://github.com/divio/django-filer/pull/1071

PR #995, opened 2017-07-14
Fix #994 (Canonical URLs computation)

### LINKED ISSUE #994: Canonical URLs computation changed between 1.26 and 1.27
Hi,

Commit 878cafe82a5d2482ca4b571f6bca8321b55261db changed the behavior of canonical URLs.

`self.uploaded_at.strftime('%s')` does not equals `int((self.uploaded_at - datetime(1970, 1, 1, tzinfo=pytz.utc)).total_seconds())`. Thus, former permalinks to filer images of files are broken.

The fix for me was to change `int((self.uploaded_at - datetime(1970, 1, 1, tzinfo=pytz.utc)).total_seconds())` to `int((self.uploaded_at - datetime(1970, 1, 1, 1, tzinfo=pytz.utc)).total_seconds())` to get the old behavior.

It is quite a serious problem for me as I have thousands of links broken.

What do you think?

Thanks

aleray:
Fixes the computation of Canonical URLs to behave as before commit 878cafe.

See issue #994

### DIFF
--- a/filer/models/filemodels.py
+++ b/filer/models/filemodels.py
@@ -282,9 +282,9 @@ class File(PolymorphicModel, mixins.IconsMixin):
     @property
     def canonical_time(self):
         if settings.USE_TZ:
-            return int((self.uploaded_at - datetime(1970, 1, 1, tzinfo=timezone.utc)).total_seconds())
+            return int((self.uploaded_at - datetime(1970, 1, 1, 1, tzinfo=timezone.utc)).total_seconds())
         else:
-            return int((self.uploaded_at - datetime(1970, 1, 1)).total_seconds())
+            return int((self.uploaded_at - datetime(1970, 1, 1, 1)).total_seconds())
 
     @property
     def canonical_url(self):


stefanfoulis:
@aleray Thanks for the PR. Maybe we should store the canonical URL in the database instead since it is not supposed to every change. Then the logic for generating the URL can be changed or customised over time without breaking existing links.
The commit that introduced the change said *Fixed canonical urls*. Do we know what bug was fixed with the change? (/cc @TheDeadOne )

TheDeadOne:
@stefanfoulis `%s` is not supported format string for `strftime`, it is not portable, it may silently produce a wrong result for an aware `datetime` object, it fails if an input is in UTC but local timezone is not UTC. The only reason it's working (on unix-like systems) is because Python is passing the information to system's `strftime`. Under Windows it does not work at all.

yakky:
@aleray  I cherry-picked commits on a sane develop branch which builds on osx (hopefully)
I'll open a separate issue because long-term, storing the canonical URL (or at least the non configured part) might be a more robust solution

PR #997, opened 2017-07-20
Add and correct pt-br translations

jlucchese:


yakky:
@jlucchese thanks for contributing this
Translation must be handled through transifex https://www.transifex.com/divio/django-filer/language/pt_BR/
Do you mind updating translation there?
merging on github would make very hard to sync them with transifex

yakky:
translations updated from transifex

PR #999, opened 2017-07-24
Drop Django<1.8 support

stefanfoulis:


filer/admin/clipboardadmin.py
```diff
@@ -75,16 +74,17 @@ def ajax_upload(request, folder_id=None):
...
+    response_params = {'content_type': mimetype}
     folder = None
     if folder_id:
         try:
             # Get folder
             folder = Folder.objects.get(pk=folder_id)
         except Folder.DoesNotExist:
-            return HttpResponse(json.dumps({'error': NO_FOLDER_ERROR}),
-                                **response_params)
+            return HttpResponse(
```
czpython: maybe JSONResponse?

stefanfoulis: Good point. I wonder if there even is any case where it would make sense to return json as ``text/html``.

PR #1000, opened 2017-07-24
Fix swappable image

stefanfoulis:
Squashed version of #857 with updated testsuite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,14 @@ after_success:
   - mkdir -p shippable/codecoverage && cp coverage.xml shippable/codecoverage/coverage.xml
 
 env:
-  - DJANGO="1.8"
-  - DJANGO="1.9"
-  - DJANGO="1.10"
-  - DJANGO="1.11"
+  - DJANGO="1.8" SWAP="no"
+  - DJANGO="1.8" SWAP="yes"
+  - DJANGO="1.9" SWAP="no"
+  - DJANGO="1.9" SWAP="yes"
+  - DJANGO="1.10" SWAP="no"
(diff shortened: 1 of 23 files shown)


stefanfoulis:
@skirsdeda I'm still getting the error with ``Django>=1.10``.

skirsdeda:
@stefanfoulis It seems that I'll have to fix django-polymorphic because it tries to validate model manager on polymorphic models (only excluding abstract models) whereas manager on a swapped model is inaccessible as well. So this might be a bit of a setback unless we can work out some workaround.

skirsdeda:
@stefanfoulis I did the tests and confirmed that the problem is in django-polymorphic. See django-polymorphic/django-polymorphic#307 which fixes the issue.

stefanfoulis:
@skirsdeda cool! 
Since this PR works for non swappable models anyway and is already an great improvement in its current state, I'd say we should merge it already now.
Then when django-polymorphic is released with that fix, we can just update the test-suite to require passing tests for ``Django>=1.10`` with swappable models.

skirsdeda:
@stefanfoulis agreed 👍

stefanfoulis:
Hmmm....I wonder why we're limiting to ``django-polymorphic<1.1`` in ``setup.py`` (https://github.com/divio/django-filer/blob/60c15fe9728c6b40a0b8cfdcb8a4b9a6f0ed3fb1/setup.py#L32)

skirsdeda:
My guess is that this limit is completely artificial because it was just raised from another version to what was the newest version of django-polymorphic at that time.

PR #1001, opened 2017-07-25
Bump dependency constraints to current releases

stefanfoulis:
Let's see what travis says to this.


skirsdeda:
@stefanfoulis Now that django-polymorphic 1.3 is out, it should be just `>=1.3`. It contains the fix for swappable models.

yakky:
Dependencies have already been updated by other PR

PR #1003, opened 2017-07-26
Attempt to fix #1002

### LINKED ISSUE #1002: [develop] automatically generated migrations require settings.FILER_IMAGE_MODEL
With #1000 any 3rd party app creating a migration that uses ``FilerImageField`` will create an entry like this in the migrations:

```
migrations.AlterField(
    model_name='article',
    name='featured_image',
    field=filer.fields.image.FilerImageField(on_delete=django.db.models.deletion.SET_NULL, verbose_name='featured image', blank=True, to=settings.FILER_IMAGE_MODEL, null=True),
),
```
And will fail to run because ``settings.FILER_IMAGE_MODEL`` is not set for most projects.
Is there a way to set a default for this value in settings?

skirsdeda:
I think this is the least invasive approach. Also some pre Django 1.8 compat code removed.

### DIFF
--- a/filer/fields/file.py
+++ b/filer/fields/file.py
@@ -138,12 +138,3 @@ class FilerFileField(models.ForeignKey):
         }
         defaults.update(kwargs)
         return super(FilerFileField, self).formfield(**defaults)
-
-    def south_field_triple(self):
-        "Returns a suitable description of this field for South."
-        # We'll just introspect ourselves, since we inherit.
-        from south.modelsinspector import introspector
-        field_class = "django.db.models.fields.related.ForeignKey"
-        args, kwargs = introspector(self)
-        # That's our definition!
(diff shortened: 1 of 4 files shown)


stefanfoulis:
@skirsdeda feels a bit hacky to set the value on the ``settings`` object. But I agree that this is probably the least invasive approach and should work well.
Is this PR ready to merge?

skirsdeda:
@stefanfoulis Yes, a bit hacky to say the least :) But this is what https://github.com/wq/django-swappable-models does for example. And while swappable models integrated into Django itself have the comfort of setting the default setting's value (such as `AUTH_USER_MODEL`), we don't get to do that. So this is the best approach in this case.

And yes, this is ready to be merged.

stefanfoulis:
@skirsdeda thanks!

PR #1005, opened 2017-08-09
Fixed `get_css_position` filter breaking when there is no image

vxsx:


vxsx:
@stefanfoulis is the image was removed from filer but there is still a plugin that references it which template uses this filter. Any value would make sense as long as it is not AttributeError

PR #1007, opened 2017-08-15
Look for explicitly defined FILER_UPLOADER_CONNECTIONS first

JamieRobertson:
In some production environments, settings are imported as modules. In
this case there is a possibility that DATABASES['default'] will not be
defined when this code is run, in those cases, a keyError will be
thrown. This an attempt to fix this edge case bug.


JamieRobertson:
Personal comment: This is my first PR for django filer. Please let me know if I have followed the guidelines correctly for contributing. :)

JamieRobertson:
@yakky OK I'm merging develop into this branch. I don't have permission to merge this into develop.

PR #1021, opened 2017-11-17
Fix File.is_public default value

skirsdeda:
Currently whenever `FILER_IS_PUBLIC_DEFAULT` is changed, `makemigrations` generates a new migration for filer's `File` model. Changing the default value to use a callable mitigates this problem.


skirsdeda:
Since adding new migrations to separate PRs is a pain, I just added additional fix here: make `File._file_size` a `BigIntegerField`. Currently, django-filer fails to upload files larger than max int (signed) in database just because of this field.

benzkji:
If your patch will not set a default on database level, is this an issue? It will have a default of `True`, because of older migrations, right? Or is the default handled "ORM" only, ie will not work at all when inserting without ORM?

skirsdeda:
@benzkji, Django model fields never save default values at database level. It's always handled by ORM.

benzkji:
@skirsdeda true, just checked my db. thanks!

benzkji:
How about merging this? Broken Travis seems to be a OSX issue (pip: command not found) and not an issue with the changes?

yakky:
@skirsdeda could I ask you to split this in two PR? I swear I will merge them promptly and in the right order :bowing_man:

skirsdeda:
@yakky, not sure why travis build is still failing but should be good.

yakky:
For some reason macos builds are failing again

yakky:
(not related to this PR, I'm thinking to merge this anyway and have a look at the build failures later)

skirsdeda:
@yakky You also added 0008 migration in #1030 so that one will have to be renumbered to 0010 and dependencies changed 0009 if these 2 are to be merged sooner.

yakky:
@skirsdeda this will surely land before the django 2.0 related ones. A rebase of those is needed anyway. Thanks for the heads up

yakky:
@skirsdeda I cherry-picked commits on a sane develop branch which builds on osx (hopefully)

yakky:
Merged with #1054

PR #1022, opened 2017-11-21
Fix RemovedInDjango20Warnings

skirsdeda:
This fixes deprecation warnings shown by Django 1.11. Compatibility with Django 1.8 is kept. There are two more deprecation warnings which require newer Django versions to be fixed so those can wait.


yakky:
@skirsdeda thanks, I included this in #1030 which provides full Django 2.0 support

skirsdeda:
Closing in favor of #1030.

Issue #973, opened 2017-04-03
Abnormally high memory consumption


leonsmith:
https://github.com/divio/django-filer/pull/820

The above PR seems to have the side effect of huge memory consumption (multiple GBs to the point the machine OOM kills the process) & the preview page timing out.

Trying to debug the issue for the root cause but setting the height param to be 0 instead seems to resolve it whilst keeping the intent of this original fix

```python
        'admin_sidebar_preview': {
            'size': (SIDEBAR_IMAGE_WIDTH, 0), 
            'upscale': True
        },
```

leonsmith:
I assume with the new upscale: True flag filer is attempting to create a 10000 high image, will update once I'v managed to allow the generation to run without it being OOM killed

leonsmith:
Confirmed this is now causing django to generate a 10000 high image coming in at 6MB and a 2x one coming in at 16.4MB from an original image size of 280K

```
4.0K	./file000132701536.jpg__16x16_q85_crop_subsampling-2_upscale.jpg
4.0K	./file000132701536.jpg__16x16_q85_crop_subsampling-2_upscale@2x.jpg
 12K	./file000132701536.jpg__180x180_q85_crop_subsampling-2_upscale.jpg
 28K	./file000132701536.jpg__180x180_q85_crop_subsampling-2_upscale@2x.jpg
5.7M	./file000132701536.jpg__210x10000_q85_crop-scale_subsampling-2_upscale.jpg
 16M	./file000132701536.jpg__210x10000_q85_crop-scale_subsampling-2_upscale@2x.jpg
4.0K	./file000132701536.jpg__32x32_q85_crop_subsampling-2_upscale.jpg
4.0K	./file000132701536.jpg__32x32_q85_crop_subsampling-2_upscale@2x.jpg
4.0K	./file000132701536.jpg__48x48_q85_crop_subsampling-2_upscale.jpg
4.0K	./file000132701536.jpg__48x48_q85_crop_subsampling-2_upscale@2x.jpg
4.0K	./file000132701536.jpg__64x64_q85_crop_subsampling-2_upscale.jpg
8.0K	./file000132701536.jpg__64x64_q85_crop_subsampling-2_upscale@2x.jpg
```

PR #974, opened 2017-04-03
Reduce target height of admin_sidebar_preview thumbnail to 0

leonsmith:
Fix for #973

PR #1024, opened 2017-11-24
Out of memory error fix.

Itsindigo:
As described by @leonsmith in issue #973, 10000 value is causing timeout and eventually killing process.


yakky:
Thanks @Itsindigo for raising this again
marking as duplicate of #974

PR #1034, opened 2018-02-02
do not override _width and _height if file is not ready yet

jrief:


jrief:
Just a not on failing unit tests:

There must be a problem with the setup, one test fails saying:

1.
```
./filer/admin/imageadmin.py:0:1: I001 isort found deviations to configured sorting rules, run it on the file to fix this.
```

but I haven't event touched that file, and I don't even see what I could do to avoid it.

2.
```
/Users/travis/.travis/job_stages: line 57: pip: command not found
```

this certainly has nothing to do with my pull request.

yakky:
@jrief would you rebase this on top of develop?
The isort failure could be caused by some different isort version, but I no longer see now that builds run fine again

filer/models/abstract.py
```diff
@@ -68,7 +68,8 @@ def file_data_changed(self, post_init=False):
                 self._width, self._height = PILImage.open(imgfile).size
                 imgfile.seek(0)
             except Exception:
-                self._width, self._height = None, None
+                if post_init is False:
```
yakky: Could you explain the rationale behing this?
Probably it would be useful to add a comment. I remember setting these has been quite controversial in the past

jrief:
@yakky rebased on top of develop.

About the changelog: This fix has been documented for version 1.2.8, although it didn't work there. Therefore, we might have to move this statement to the unreleased version 1.3.1.

yakky:
@jrief Could you handle this in this PR? So we don't miss it

yakky:
IDK why tests are failing again

filer/models/abstract.py
jrief: done

jrief:
@yakky fixed all your annotations

yakky:
@jrief  I cherry-picked commits on a sane develop branch which builds on osx (hopefully)

yakky:
Merged with #1052 
Thanks @jrief

Issue #966, opened 2017-02-22
import_files command not compatible with Django 1.10


2martens:
The following code is even existing in your most recent commit here on GitHub.

```python
class Command(NoArgsCommand):
    """
    Import directory structure into the filer ::

        manage.py --path=/tmp/assets/images
        manage.py --path=/tmp/assets/news --folder=images
    """

    option_list = BaseCommand.option_list + (
        make_option('--path',
            action='store',
            dest='path',
            default=False,
            help='Import files located in the path into django-filer'),
        make_option('--folder',
            action='store',
            dest='base_folder',
            default=False,
            help='Specify the destination folder in which the directory structure should be imported'),
    )
```
Multiple things are wrong here. First there is no NoArgsCommand existing. It has to be BaseCommand. Second there is no option_list in BaseCommand. Instead there is a method called add_arguments which gets a parser as parameter. You need to overwrite this method. A possible solution is shown below.

```python
class Command(BaseCommand):
    """
    Import directory structure into the filer ::

        manage.py --path=/tmp/assets/images
        manage.py --path=/tmp/assets/news --folder=images
    """
    def add_arguments(self, parser):
        parser.add_argument('--path',
            action='store',
            dest='path',
            default=False,
            help='Import files located in the path into django-filer')
        parser.add_argument('--folder',
            action='store',
            dest='base_folder',
            default=False,
            help='Specify the destination folder in which the directory structure should be imported')
```

yakky:
Fixed by #1058

PR #1058, opened 2018-04-09
Feature/fix import files command

jdcookie:
Update the import_files command so that is does not use deprecated methods and classes.
Works with Django 1.10+
A simple edit to resolve Issue #966

### DIFF
--- a/filer/management/commands/import_files.py
+++ b/filer/management/commands/import_files.py
@@ -2,7 +2,6 @@
 from __future__ import absolute_import, unicode_literals
 
 import os
-from optparse import make_option
 
 from django.core.files import File as DjangoFile
 from django.core.management.base import BaseCommand
--- a/filer/management/commands/import_files.py
+++ b/filer/management/commands/import_files.py
@@ -119,7 +119,7 @@ class Command(BaseCommand):
             '--path',
             action='store',
             dest='path',
-            default=False,
+            required=True,
             help='Import files located in the path into django-filer'
         )
 
--- a/filer/management/commands/import_files.py
+++ b/filer/management/commands/import_files.py
@@ -80,7 +80,7 @@ class FileImporter(object):
         This method walk a directory structure and create the
         Folders and Files as they appear.
         """
-        path = path or self.path if self.path is not False else ''
+        path = path or self.path or ''
         base_folder = base_folder or self.base_folder
         # prevent trailing slashes and other inconsistencies on path.
         path = os.path.normpath(upath(path))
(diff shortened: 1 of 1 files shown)


skirsdeda:
@yakky So maybe I should revive #500? It would also make importfiles management command more useful.

yakky:
@skirsdeda my plan is actually to review both
I would like to keep 1.3.1 "easy", and thus including this to make the command at least work across all the django version.
I think that #500 is more suitable for release 1.4, which I think we could release shortly (just dropping older django and including the PR already approved / evaluated for a shorter release cycle)

skirsdeda:
@yakky sounds good.

filer/management/commands/import_files.py
```diff
@@ -81,7 +80,7 @@ def walker(self, path=None, base_folder=None):
         This method walk a directory structure and create the
         Folders and Files as they appear.
         """
-        path = path or self.path
+        path = path or self.path if self.path is not False else ''
```
yakky: path must be a string in any case, I think we should rewrite this as `path or self.path or ''`

jdcookie: ok - will update it now.

filer/management/commands/import_files.py
```diff
@@ -108,27 +107,30 @@ def walker(self, path=None, base_folder=None):
...
     """
-
-    option_list = BaseCommand.option_list + (
-        make_option('--path',
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--path',
             action='store',
             dest='path',
             default=False,
```
yakky: Can I ask you to mark path as required? This https://github.com/divio/django-filer/pull/1058/files#diff-361fe6a032f5ff57a38ea06ab3c367a0R83 fixes an issue if path is empty, but it opens the door to an undesired behavior (importing the whole django project root)

jdcookie: Should we drop the default then, if its required no need to have a default value?

yakky: Yes, we can remove it, as it's redundant

jdcookie:
Should be working now
```
root@9e22bc1e495a:/usr/src/app# ./manage.py import_files
usage: manage.py import_files [-h] [--version] [-v {0,1,2,3}]
                              [--settings SETTINGS] [--pythonpath PYTHONPATH]
                              [--traceback] [--no-color] --path PATH
                              [--folder BASE_FOLDER]
manage.py import_files: error: the following arguments are required: --path
```

yakky:
Thanks @jdcookie !
Waiting for tests to finish to merge

yakky:
Merged, thanks again for working on this

jdcookie:
No worries

PR #1059, opened 2018-04-09
Sync i18n from transifex

yakky:
It's large enough that I prefer to give it a run on travis before merge


PeterW-LWL:
The german translation is updated on transifex.

yakky:
@PeterW-LWL thanks, PR updated

yakky:
Merging this now. Updates will be pulled shortly before release

PR #1061, opened 2018-04-11
use self.get_queryset in folderadmin instead of Folder.objects

skirsdeda:
This is to support better customizability in folderadmin. I was planning to go further with this, but got tangled up with other work. But it can be merged as-is, I think.


yakky:
Thanks @skirsdeda !

PR #1062, opened 2018-04-11
allow django_polymorphic==2.0

skirsdeda:
Thing is, that polymorphic 1.3 has nasty bug when swapped models are used (like swapped Image model in filer). See django-polymorphic/django-polymorphic@02330a5ec835b118d67aacf071cc2deef1250829. Unfortunately, the fix is only included in django-polymorphic 2.0 which only supports Django 1.11 and newer. The problem first appeared in Django 1.10, so it is impossible to have polymorphic swapped model with released version of django-polymorphic.
Therefore, for people using swapped image model and Django 1.11, django-polymorphic 2.0 is required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,39 +131,23 @@ matrix:
       python: "3.6"
       env: DJANGO="master" SWAP="yes"
 
-# image model swapping is failing in Django>=1.10
+# image model swapping is failing in Django==1.10 because django_polymorphic 2.0
+# only support Django>=1.11
     - os: linux
       python: "2.7"
       env: DJANGO="1.10" SWAP="yes"
     - os: linux
-      python: "2.7"
-      env: DJANGO="1.11" SWAP="yes"
-    - os: linux
       python: "3.4"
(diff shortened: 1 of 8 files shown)


skirsdeda:
Updated test requirements and test matrix accordingly.

yakky:
@skirsdeda this is super nice! Could you add a not in the documentation to reflect this?

skirsdeda:
@yakky sure, will do.

skirsdeda:
Updated docs. Didn't write anything about Django 1.10 situation, maybe we can sort it out meanwhile. I think the best option would be to have django-polymorphic support release with that small fix. I asked about that so we'll see how it goes.

yakky:
@skirsdeda :+1: I will check during the weekend if I can fix this into filer in time for 1.3.1

yakky:
@skirsdeda this looks good and I am going to merge it
I'll try to come up with something for 1.10
I will release 1.3.1 anyway during the weekend, because it's going to fix the situation for a lot of cases anyway
Thanks for your work on this!

PR #1064, opened 2018-04-16
Use django_polymorphic 1.3.1 to solve swapped model bug

skirsdeda:


yakky:
@skirsdeda thanks! As the only user-facing change is the documentation one, which will be updated anyway, I don't think this worth a new release. What do you think?

skirsdeda:
@yakky yeah, let's leave this for 1.4.

PR #1066, opened 2018-04-23
unlimit dropzone filesize

skirsdeda:
Since filer is using older dropzone, it defaults to max file size of 256 MB. This is way too low and there's no reasons to limit it from django side.

### DIFF
--- a/filer/static/filer/js/addons/dropzone.init.js
+++ b/filer/static/filer/js/addons/dropzone.init.js
@@ -24,7 +24,6 @@ django.jQuery(function ($) {
     var hiddenClass = 'hidden';
     var mobileClass = 'filer-dropzone-mobile';
     var objectAttachedClass = 'js-object-attached';
-    // var dataMaxFileSize = 'max-file-size';
     var minWidth = 500;
     var checkMinWidth = function (element) {
         element.toggleClass(mobileClass, element.width() < minWidth);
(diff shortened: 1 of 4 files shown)


filer/static/filer/js/addons/dropzone.init.js
```diff
@@ -65,8 +64,8 @@ django.jQuery(function ($) {
             url: dropzoneUrl,
             paramName: 'file',
             maxFiles: 1,
-            // for now disabled as we don't have the correct file size limit
-            // maxFilesize: dropzone.data(dataMaxFileSize) || 20, // MB
+            // filer and django do not limit file size, let's use some ridiculously large value
+            maxFilesize: 4000000, // MB
```
FinalAngel: can you set this to ``null``? Should do the same according to their docs

filer/static/filer/js/addons/table-dropzone.js
```diff
@@ -74,8 +74,8 @@
                     url: dropzoneUrl,
                     paramName: 'file',
                     maxFiles: 100,
-                    // for now disabled as we don't have the correct file size limit
-                    // maxFilesize: dropzone.data(dataMaxFileSize) || 20, // MB
+                    // filer and django do not limit file size, let's use some ridiculously large value
+                    maxFilesize: 4000000, // MB
```
FinalAngel: Same here

FinalAngel:
@skirsdeda thanks for the contribution! Could you check the comments I made? That'd be awesome.

PR #1068, opened 2018-04-26
Inherit Folder from MPTTModel instead of using mptt.register().

sir-sigurd:
The preferred way to do model registration is via model inheritance and `mptt.AlreadyRegistered` is deprecated since django-mptt 0.4

https://django-mptt.readthedocs.io/en/latest/upgrade.html?highlight=mptt.register#use-model-inheritance-where-possible


yakky:
@jrief @FinalAngel how about merging this?

PR #1076, opened 2018-05-24
Add thumbnail view for folder list admin view

0mk1:


filer/templates/admin/filer/folder/directory_thumbnail_list.html
```diff
@@ -68,7 +68,7 @@
           <p>Files</p>
```
vxsx: here as well

PR #1077, opened 2018-05-24
Fix #1035 -- Hide 'Save as new' button

### LINKED ISSUE #1035: hide "Save as New" button
A click on "Save as New" will raise into PermissionDenied, because filer.admin.permissions.PrimitivePermissionAwareModelAdmin.has_add_permission() will always return False, see: https://github.com/divio/django-filer/blob/develop/filer/admin/permissions.py#L9

So, the "Save as New" button should not be exists...

joncasdam:
Remove the hack (as defined by the code comment) so the button is not
visible anymore.

Also, it seems to me that the 'hack' is not needed anymore - if it's still needed, this PR might introduce breaking changes.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -6,6 +6,7 @@ CHANGELOG
 
 * Enabled django-mptt 0.9
 * Converted QueryDict to dict before manipulating in admin
+* Hide 'Save as new' button in file admin
 
 1.3.2 (2018-04-23)
 ------------------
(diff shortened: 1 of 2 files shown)


yakky:
@joncasdam would you add a line in the changelog? I already opened the 1.3.3 section

joncasdam:
Sure! On it.

joncasdam:
Done: amended.

yakky:
@joncasdam thanks for your contribution!

PR #1078, opened 2018-05-25
Raises Unidecode package dependency to v1.0

wiesion:
Needed to raise this dependency for a project and it seems to be working fine - unfortunately, i have no time at hand right now to set up the tox environment properly (it fails at a lot of stuff) so i can't say if it passes tests (sorry for that, wish i had more time).


yakky:
@wiesion thanks for this. It's definitely about time to allow newer versions in
I would keep the lower bound unchanged if there is no specific reason

PR #1079, opened 2018-05-26
Allow mptt 0.9

yakky:
Fix #1070

PR #1080, opened 2018-05-27
Reenable missing tests for custom image models

yakky:
Some tests have been disabled for extended images
They can be safely re-enabled now along with some further fixes for extended images

#1079 must be applied first to allow tests to run here

### DIFF
--- a/filer/admin/imageadmin.py
+++ b/filer/admin/imageadmin.py
@@ -90,11 +90,17 @@ class ImageAdminForm(forms.ModelForm):
 
 
 class ImageAdmin(FileAdmin):
+    change_form_template = 'admin/filer/image/change_form.html'
     form = ImageAdminForm
 
 
+if FILER_IMAGE_MODEL == 'filer.Image':
+    extra_main_fields = ('author', 'default_alt_text', 'default_caption',)
+else:
+    extra_main_fields = ('default_alt_text', 'default_caption',)
+
 ImageAdmin.fieldsets = ImageAdmin.build_fieldsets(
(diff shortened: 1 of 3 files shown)


filer/admin/imageadmin.py
```diff
@@ -90,11 +90,17 @@ class Media(object):
 
 
 class ImageAdmin(FileAdmin):
+    change_form_template = 'admin/filer/image/change_form.html'
     form = ImageAdminForm
 
 
+if FILER_IMAGE_MODEL == 'filer.Image':
+    extra_main_fields = ('author', 'default_alt_text', 'default_caption',),
```
czpython: looks like there's an extra comma at the end

filer/admin/imageadmin.py
```diff
@@ -90,11 +90,17 @@ class Media(object):
...
+else:
+    extra_main_fields = ('default_alt_text', 'default_caption',),
```
czpython: shouldn't we do a try/except to check if author field is present, instead of hardcoding check for image model?

yakky: Here we really want to check if we are using the "original" filer model; In the end it doesn't really matter: if I extend the image model I will redefine the admin anyway
But in case you want to check field instead, I suggest using `if hasattr` instead

filer/admin/imageadmin.py
yakky: Fixed

yakky:
@czpython 1 comment fixed, see the reply https://github.com/divio/django-filer/pull/1080#discussion_r191830203 to the second one

Issue #1075, opened 2018-05-18
AttributeError: This QueryDict instance is immutable  in _set_previous_subject_location


mireq:
Line

`self.data['subject_location'] = subject_location`

in filer.admin.imageadmin raises exception "QueryDict instance is immutable". Attribute self.data should be either a copy of POST (self.data = self.data.copy()) or should be changed to mutable (self.data._mutable = True).

yakky:
Confirmed, even if only in case the subject location is wrong.
But worth a fix anyway

yakky:
Fixed in #1081

wfehr:
I am still getting this error with following scenario:
- upload image with 10.000x10.000px
- subject location is ~5.000x5.000px
- replace the image with an image with 2.000x2.000px
- subject location is still 5.000x5.000px
-> click SAVE -> internal server error

```
End of trace:
File "/usr/local/lib/python3.6/site-packages/filer/admin/imageadmin.py" in clean_subject_location
  69.         self._set_previous_subject_location(cleaned_data)

File "/usr/local/lib/python3.6/site-packages/filer/admin/imageadmin.py" in _set_previous_subject_location
  38.         self.data['subject_location'] = subject_location

File "/usr/local/lib/python3.6/site-packages/django/http/request.py" in __setitem__
  434.         self._assert_mutable()

File "/usr/local/lib/python3.6/site-packages/django/http/request.py" in _assert_mutable
  431.             raise AttributeError("This QueryDict instance is immutable")`
```

The subject location should get deleted when the image "in the background" gets changed.
Are you able to recreate this?
I have installed:
- django==1.11.12
- django-cms==3.5.2
- django-filer==1.3.1

misli:
I have also caught this error in Sentry.
Filer version: `1.3.2`
Request: `POST /admin/filer/image/{var}/change/`
End of traceback:
```
  File "filer/admin/imageadmin.py", line 69, in clean_subject_location
    self._set_previous_subject_location(cleaned_data)
  File "filer/admin/imageadmin.py", line 38, in _set_previous_subject_location
    self.data['subject_location'] = subject_location
  File "django/http/request.py", line 434, in __setitem__
    self._assert_mutable()
  File "django/http/request.py", line 431, in _assert_mutable
    raise AttributeError("This QueryDict instance is immutable")
```

PR #1081, opened 2018-05-27
Convert QueryDict to dict before manipulating in admin

yakky:
#1075 

Test coverage provided by #1080
This must be merged before #1080 to allow tests there to run


yakky:
Thanks @czpython

Issue #1006, opened 2017-08-14
Secure downloads view should be decorated with @never_cache


irsalmashhor:
When trying to view a filer secure file, it looks like it is trying to add this to the cache and causing an unrelated memcache limit error, as the django cache middleware is trying to the result of this view into the cache.

It doesn't seem like these views should be cached at all, unless I'm missing something - https://github.com/divio/django-filer/blob/develop/filer/server/views.py

PR #1100, opened 2018-10-24
Added never_cache decorator in server views.

davideferre:
Solving issue #1006


vThaian:
@czpython could you look at it?

Issue #1109, opened 2018-11-20
1.4.0 distfile contains .DS_Store


moubctez:
Please, remove filer/static/.DS_Store from the distribution archive. :)

jrief:
confirmed.

PR #1110, opened 2018-11-20
Prevents issues such as #1109

jrief:


jrief:
Please restart job #2238.53 on Travis-CI

Issue #1125, opened 2019-01-15
filer/fields/file.py raises error when fields are empty


debaze:
In the Admin change view, When I try to save an empty file field, I get an error:

    invalid literal for int() with base 10: ''
    /django/db/models/fields/__init__.py in get_prep_value, line 965

From the traceback, I get it is caused by submitting an empty string to a query in the `file.py`

```
site-packages/filer/fields/file.py in render
28.         obj = self.obj_for_value(value) 
```
Variable | Value
----- |----
__class__ | <class 'filer.fields.file.AdminFileWidget'>
attrs | {'id': 'id_image'}
name | 'image'
renderer | None
self | <filer.fields.image.AdminImageWidget object at 0x7fee4c164b70>
value | ''

I could hack a solution together by checking if `value == ''` before trying to retrieve the object in `obj_for_value`, but I'm not sure if that is the best way to fix it..

Versions:

Package | Name | Version
-- | -- | --
Django |   | 2.1
Python |   | 3.6.6
ckeditor | Ckeditor | 5.6.1
debug_toolbar | Debug Toolbar | 1.11
django_extensions | Django_Extensions | 2.1.4
easy_thumbnails | Easy_Thumbnails | <module 'easy_thumbnails.get_version' from '---/site-packages/easy_thumbnails/get_version.py'>
filer | Filer | 1.4.3

PR #1127, opened 2019-01-15
do not retrieve object for empty values

debaze:
https://github.com/divio/django-filer/issues/1125


vThaian:
Thanks, @debaze for you PR. I created another one based on your #1126.

PR #1130, opened 2019-01-22
Fixed #1129 - missing validation for folder plugin when folder field are empty

### LINKED ISSUE #1129: Folder plugin raises error when folder field are empty
An issue is similar to #1125

When I'm trying to save Folder plugin on my test page without selected folder then I'm getting an error:
```
Exception Type: ValueError at /en/admin/cms/page/add-plugin/
Exception Value: invalid literal for int() with base 10: ''
```

Django: 2.1
Filer: 1.4.3 and 1.4.4
Python: 3.5

vThaian:
Based on issue #1129

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+1.4.5 (unreleased)
+------------------
+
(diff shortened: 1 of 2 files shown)


FinalAngel:
Fixed in #1165

PR #1139, opened 2019-03-21
add persian language support

katoozi:
translate messages for persian language


katoozi:
i think the flake8 have a issue, because i did't changed python code.

Issue #460, opened 2014-11-04
SVG Image Support - Proposal


ghost:
Is it possible to add support for SVG images?

gavinwahl:
What need to be done for SVG support?

yakky:
The main issue is how to handle thumbnailing. SVG must be handled by the Image model as they can be used anywhere an "normal" image is used. But thumbnailing has no meaning for SVGs, thus we we must come up with a strategy to handle it in the templates

aleray:
SVG support would be very helpfull to me as well.

Thanks,

Alex

jrief:
@yakky so why not bypassing the thumbnailing step? That would be quite easy to implement.
The remaining problem is clipping, but in my opinion this should be handled by the browser.

nilsborchers:
:+1: That would be suuper helpful actually!

nilssolanki:
Yes, I would really appreciate SVG support as well! I found the following plugin, but it hasn't been updated in three years: https://github.com/agcooke/django-filer-svg

lmanzurv:
+1

quater:
I just stumbled over this too. It would be great to have *.svg support.

As of my research I figured that Nick Khlestov has implemented this with Django generically and I believe this might be a viable way of doing this with django-filer?

**Detailed blog post:**
https://7webpages.com/blog/how-to-have-svg-allowed-as-an-image-for-django-rest-framework/

**Code for SVGAndImageFormField class:**
https://gist.github.com/ambivalentno/9bc42b9a417677d96a21

**Related discussion on Stackoverflow:**
http://stackoverflow.com/questions/38006200/allow-svg-files-to-be-uploaded-to-imagefield-via-django-admin/38010444#38010444

benzkji:
Any news on this? It bothers me to make my image fields `FilerFileFields`, just because one could want to insert an svg.

yakky:
We don't have enough resources to tackle this right now, but if anyone wants to step up and start something, I'm more than willing to help and review the code

benzkji:
one thing is having a SVG in a FilerImageField. Another thing is how to deal with them in all those {% thumbnail object.image 100x100 %} template tags...those will probably all fail, as easy_thumbnails is not made for creating thumbs from SVGs (what makes no sense anyway). creating a {% thumbnail_or_svg ... %} wrapper could be an option...but no clue if this should go into the filer codebase...!?

lmanzurv:
[...]

PR #1143, opened 2019-04-25
bypassing the thumbnailing step for specific file #460

haricot:
related #460
Give the ability to download any chosen file extension rather than just files where thumbnail creation is possible on the fly.
With this we can add SVG files and we see the thumbnail  if size <=4ko , if not the file has not preview.

### DIFF
--- a/filer/settings.py
+++ b/filer/settings.py
@@ -254,4 +254,5 @@ FILER_DUMP_PAYLOAD = getattr(settings, 'FILER_DUMP_PAYLOAD', False)  # Whether t
 
 FILER_CANONICAL_URL = getattr(settings, 'FILER_CANONICAL_URL', 'canonical/')
 
-FILER_FILE_EXTENSION_NOTHUMBS =  getattr(settings, 'FILER_FILE_EXTENSION_NOTHUMBS', '.svg','.gltf' )
+FILER_FILE_EXTENSION_ACCEPTED = ['.jpg', '.jpeg', '.png', '.gif', '.svg', ,'.gltf']
+FILER_FILE_EXTENSION_NOTHUMBS =  getattr(settings, 'FILER_FILE_EXTENSION_NOTHUMBS', ['.svg' ,'.gltf']   )
--- a/filer/settings.py
+++ b/filer/settings.py
@@ -253,3 +253,5 @@ FILER_UPLOADER_CONNECTIONS = getattr(
 FILER_DUMP_PAYLOAD = getattr(settings, 'FILER_DUMP_PAYLOAD', False)  # Whether the filer shall dump the files payload
 
 FILER_CANONICAL_URL = getattr(settings, 'FILER_CANONICAL_URL', 'canonical/')
+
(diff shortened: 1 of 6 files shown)


jrief:
This is a feature, I'm awaiting for a long time. Thanks!

filer/settings.py
```diff
@@ -253,3 +253,6 @@ def update_server_settings(settings, defaults, s, t):
 FILER_DUMP_PAYLOAD = getattr(settings, 'FILER_DUMP_PAYLOAD', False)  # Whether the filer shall dump the files payload
 
 FILER_CANONICAL_URL = getattr(settings, 'FILER_CANONICAL_URL', 'canonical/')
+
+FILER_FILE_EXTENSION_ACCEPTED = ['.jpg', '.jpeg', '.png', '.gif', '.svg']
+FILER_FILE_EXTENSION_NOTHUMBS = getattr(settings, 'FILER_FILE_EXTENSION_NOTHUMBS', ['.svg'])
```
jrief: Just a proposal:
Why not using the MIME-type instead of the file extension to distinguish thumbnailable files.
In my optionion, the MIME-type is the canonical and unique representation of a file type and it can be easily determined in Python.

haricot:
For the format gtlf, I remove because it does not use tag `` and we can already do this by following the documentation [extending_filer](https://django-filer.readthedocs.io/en/latest/extending_filer.html).
For the svg format, PIL does not support thumbnail creation of vector files and in the future a binding with the[ resvg](https://github.com/RazrFalcon/resvg) lib may be a good option. This PR can be useful pending possibility but via [extending_filer](https://django-filer.readthedocs.io/en/latest/extending_filer.html) this is also possible but coming out of the basic model.
I wonder this if I have to close this pull request and create two specific plugins.

filer/settings.py
haricot: We could use [puremagic](https://pypi.org/project/puremagic/) (Pure python implementation of identifying files based off their magic numbers)

jrief: Another approach would be to extend the `filer.models.filemodels.File` and add an extra field `mime_type`. This field then would have to be filled after uploading a file using the HTTP request header `Content-Type: ...`.
This in my opinion would be the cleanest solution, but requires 
* an approved API change request from a core maintainer of django-filer.
* a migration path, to fill this new field for all existing installations. Here we could use [puremagic](https://pypi.org/project/puremagic/).

@FinalAngel if you approve this proposal, I could implement this feature.

jonathan-s:
@haricot Hi there, 
I'd be pretty keen to see this in django-filer. I've tested this out locally and for me it doesn't yet work to upload a svg. When I try to upload an svg I get a 500 error when trying to view the list view. It looks the error happens when trying to generate a thumbnail. 

Worth mentioning is that I've tested this on the latest djangocms 4.0. 

Also it would be good to have tests for this too!

haricot:
@jonathan-s, It's still a proof concept for now, tested on django-cms==3.6.
We need to revisit the logic using @jrief 's PR #1152.

haricot:
Close in favor of #1166

PR #1152, opened 2019-06-25
Store mimetype of uploaded file in database

jrief:
This is a proof of concept.
By storing the MIME-type in model `File`, many existing problems can be avoided:
While serving files through the `DefaultServer`, first the file had to be opened to guess the proper MIME-type, so that the HTTP-header `Content-Type` could be set. This now is not required anymore.


haricot:
With this feature, I will be able to do better PR #1143.
Thank you @ Jacob!

jrief:
**To the maintainers of django-filer: This feature is ready for review.**

It will help to fix some other problems, I encountered, while implementing. For instance, the NGiNX server until now, just swallowed the MIME-type.

From my humble point of view, the MIME-type is a really important information of a file. And it should be stored as part of the files meta-data inside the database. Having to always open the file on disk, just to find the proper MIME-type, is not such a good idea.

jrief:
This pull-request has been rebased against the latest commit on branch master, aka version 1.6.

I also have added all sorts of unit tests to check that the mime-type of uploaded files is used and checked. It would be great to see this feature it in the upcoming version 1.7.

This feature would especially make sense for the urgently required bypassing of thumbnailing SVG images as demonstrated in #1166.

filer/models/filemodels.py
```diff
@@ -91,10 +99,17 @@ class File(PolymorphicModel, mixins.IconsMixin):
...
+        help_text='MIME type of uploaded content',
+        validators=[mimetype_validator],
+        default='application/octet-stream',
+    )
+
     objects = FileManager()
 
     @classmethod
-    def matches_file_type(cls, iname, ifile, request):
+    def matches_file_type(cls, iname, ifile, mime_type):
```
yakky: This is part of the public and documented API and we need to advertise that people extending the base class (https://django-filer.readthedocs.io/en/latest/extending_filer.html#the-very-basics) will have to update the method in the subclasses

filer/migrations/0012_file_mime_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations, models
+
+import filer.models.filemodels
+
+
+def guess_mimetypes(apps, schema_editor):
+    FileModel = apps.get_model('filer', 'File')
+    for fileobj in FileModel.objects.all():
+        fileobj.mime_type, _ = mimetypes.guess_type(fileobj.file.path)
```
yakky: Just telling from the top of my head, I need to check this better, but this *might* not work on all the storage engines

jrief: > mimetypes.guess_type(url, strict=True)
Guess the type of a file based on its filename or URL, given by url. The return value is a tuple (type, encoding) where type is None if the type can’t be guessed (missing or unknown suffix) or a string of the form 'type/subtype', usable for a MIME content-type header.

According to the Python docs only the file name is used to guess the mime-type, rather than opening the file and using a magic header to determine the type. Therefore this should also work on S3BOTO.

yakky: Apparently not

``` python
>>> import mimetypes
>>> from filer.models import File
>>> file_obj = File.objects.first()
>>> print(file_obj.file.path)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/ubuntu/virtualenvs/project/lib/python3.7/site-packages/django/db/models/fields/files.py", line 57, in path
    return self.storage.path(self.name)
  File "/home/ubuntu/virtualenvs/project/lib/python3.7/site-packages/django/core/files/storage.py", line 109, in path
    raise NotImplementedError("This backend doesn't support absolute paths.")
NotImplementedError: This backend doesn't support absolute paths.
>>> mimetypes.guess_type(file_obj.file.path)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/ubuntu/virtualenvs/project/lib/python3.7/site-packages/django/db/models/fields/files.py", line 57, in path
    return self.storage.path(self.name)
  File "/home/ubuntu/virtualenvs/project/lib/python3.7/site-packages/django/core/files/storage.py", line 109, in path
    raise NotImplementedError("This backend doesn't support absolute paths.")
NotImplementedError: This backend doesn't support absolute paths.
```

yakky: but `url` will do:

``` python
>>> mimetypes.guess_type(file_obj.file.url)
('image/png', None)
```

filer/models/filemodels.py
yakky: do we want to remove `request`? I'd keep it

jrief: I don't see any advantage keeping it. It is never used and I can hardly see any possible future use-case, where the `request` objects could be needed. So in my opinion it's better to keep the signature slim.

yakky (approved):

docs/extending_filer.rst
```diff
@@ -47,11 +47,12 @@ So let's add a ``matches_file_type()`` method to the ``Video`` model:
...
-        # the extensions we'll recognise for this file type
-        filename_extensions = ['.dv', '.mov', '.mp4', '.avi', '.wmv',]
-        ext = os.path.splitext(iname)[1].lower()
-        return ext in filename_extensions
+    def matches_file_type(cls, iname, ifile, mime_type):
+        video_types = ['application/vnd.dvb.ait', 'video/x-sgi-movie', 'video/mp4', 'video/mpeg',
+                       'video/x-msvideo', 'video/x-ms-wmv', 'video/ogg', 'video/webm', 'video/quicktime']
+        return mime_type in video_types
+
+.. note:: The signature of this classmethod changed in version 1.7.
```
yakky, line 55: Just adding as a comment as it's easy to fix after merge. This is not going to be released in 1.7 but hopefully in 2.0, so we need to change the wording of this note

jrief: shall I do it, or you want to do it?

PR #1160, opened 2019-08-28
AdminFolderWidget render fixed obj_for_value function: if value equal  "" (empty string)

NMelis:
AdminFolderWidget render fixed obj_for_value function: if value equal  "" (empty string)


FinalAngel:
Fixed in #1161. Thanks for the contribution @NMelis !

PR #1165, opened 2019-10-29
Fixes value error when folder is empty

FinalAngel:
Also makes sure the spacing is correct between button and text:

image

Fixes #1165

### DIFF
--- a/filer/fields/folder.py
+++ b/filer/fields/folder.py
@@ -74,12 +74,12 @@ class AdminFolderWidget(ForeignKeyRawIdWidget):
         return '&nbsp;<strong>%s</strong>' % truncate_words(obj, 14)
 
     def obj_for_value(self, value):
+        if not value:
+            return None
         try:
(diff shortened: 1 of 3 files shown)


koniiiik:
Hmm, wouldn't this deserve a test? I also wonder if it wouldn't be better to handle this case inside `obj_for_value`, since that can already return None in certain cases.

filer/fields/folder.py
```diff
@@ -84,6 +79,8 @@ def obj_for_value(self, value):
             obj = self.rel.model._default_manager.get(**{key: value})
         except ObjectDoesNotExist:
             obj = None
+        except ValueError:
+            obj = None
```
koniiiik: Hmmm, this feels a bit dangerous to me – you're catching a very generic error from a pretty deep call tree (the entire ORM machinery), which might suppress something that you don't want to suppress. I think it would be better to explicitly handle this case (with no value) before calling into the ORM.

PR #1171, opened 2019-12-18
Allow latest unidecode version

ashishnitinpatil:
Resolves dependency issue in other projects where Unidecode >= 1.1 is needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ REQUIREMENTS = [
     'django-mptt>=0.6,<1.0',  # the exact version depends on Django
     'django_polymorphic>=0.7,<2.1',
     'easy-thumbnails>=2,<3.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -10,7 +10,7 @@
     'django-mptt>=0.6,<1.0',  # the exact version depends on Django
     'django_polymorphic>=0.7,<2.1',
     'easy-thumbnails>=2,<3.0',
-    'Unidecode>=0.04,<1.1',
+    'Unidecode>=0.04,<=1.1.1',
```
jrief: wouldn't `Unidecode>=0.04,<1.2` make more sense?

ashishnitinpatil: Yes, I hurriedly made the change for my own repo & my own requirements. <1.2 is better.

PR #1178, opened 2020-02-01
Remove Python-2 legacy

jrief:
Since the feature branch django-3.x drops support for Python-2, we can simplify the code and remove all compatibility tweaks, such as `six`, `from __future__ import absolute_import` and unicode literals.


jrief:
scheduled for 1.8.0

yakky:
@jrief I would say we can merge this and start working on django 3.1 compatibility

PR #1180, opened 2020-03-08
docs: Fix simple typo, recursivelly -> recursively

timgates42:
There is a small typo in filer/admin/folderadmin.py.

Should read `recursively` rather than `recursivelly`.


jrief:
Thanks

PR #1189, opened 2020-04-21
fixes #1188 loading jquery.js after jquery.min.js had been loaded

### LINKED ISSUE #1188: JQuery loaded by File polluting the global namespace when DEBUG = False
Encountered as a problem while using Mezzanine.

This was a regression introduced in b66a42b. Basically it causes jquery.js to be loaded after jquery.init.js when not in DEBUG.

Two-line fix, PR incoming.

mrlowe:
Because filer was requesting jquery.js in non-DEBUG instead of jquery.min.js (like django does), both were being included in the Media output. jquery.js was being included after jquery.min.js and more importantly django's jquery.init.js, so it was polluting the global namespace. This was observed in Django 1.11 and may not be a problem on later versions.

The fix simply uses jquery.min.js in non-debug mode.

### DIFF
--- a/filer/fields/file.py
+++ b/filer/fields/file.py
@@ -5,6 +5,7 @@ import logging
 import warnings
 
 from django import forms
+from django.conf import settings
 from django.contrib.admin.sites import site
 from django.contrib.admin.widgets import ForeignKeyRawIdWidget
 from django.core.exceptions import ObjectDoesNotExist
@@ -84,13 +85,14 @@ class AdminFileWidget(ForeignKeyRawIdWidget):
         return obj
 
     class Media(object):
+        extra = '' if settings.DEBUG else '.min'
         css = {
             'all': [
                 'filer/css/admin_filer.css',
             ]
         }
         js = (
-            'admin/js/vendor/jquery/jquery.js',
+            'admin/js/vendor/jquery/jquery%s.js' % extra,
             'admin/js/jquery.init.js',
             'filer/js/libs/dropzone.min.js',
             'filer/js/addons/dropzone.init.js',


jrief:
According to the Django documentation

> [Any links to the file in the codebase should point to the compressed version.](https://docs.djangoproject.com/en/3.0/internals/contributing/writing-code/javascript/#javascript-patches)

you therefore should just include the minimized version.

Personal Annotation:
I usually add [these two alternative finders](https://github.com/awesto/cookiecutter-django-shop/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.app_name%7D%7D/finders.py) to my projects and replace the default finders in my settings.py with:
```python
STATICFILES_FINDERS = [
    'project.finders.FileSystemFinder',
    'project.finders.AppDirectoriesFinder',
]
```
In DEBUG-mode, they load the expanded version of *.css and *.js files if they exist. In my opinion this is a better approach, rather than doing it on a file-per-file base.

mrlowe:
@jrief true, that's what the docs say, but this is exactly the way the django source does it up to the current version: https://github.com/django/django/blob/stable/3.0.x/django/contrib/admin/options.py#L632. Consequently, if we just emit 'jquery.min.js', it won't match what Django is emitting when DEBUG == True, and we'll get an extra jquery object in the global namespace in debug instead of in prod.

jrief:
I'm against merging this.

From reading #1188, It appears to fix a problem caused by an incompatibility with Django-1.11, which is EOL anyway. I agree, that version 1.7 should fix the `REQUIREMENTS` in `setup.py` and not support Django-1.11.

jrief:
Ups, I just had another look at the Django code. Now I understand the problem. Django itself adds the `.min` in DEBUG mode and therefore the sorting doesn't work anymore.

 This means that I now am in favor of merging this.

PR #1190, opened 2020-04-29
Fix #1183: Upload files into latest used folder

### LINKED ISSUE #1183: By default, upload files into last used folder
Currently, if someone uploads a file using drag and drop on a widget, it is stored inside the special folder named "Unsorted Files". In large installations, this gives a bad user experience. One would expect such a file in the last used folder, which in django-filer is handled by the session variable `filer_last_folder_id`.

By adding one line in `filer.admin.clipboardadmin.ajax_upload()`, this could be fixed.

Please comment on this feature proposal.

jrief:


jrief:
This PR fixes #1183 which is a nasty usability bug.

PR #1194, opened 2020-06-24
Fix missing required atribute "on_delete" in FilerFolderField.

zbohm:
Parameter ``on_delete`` in ``__init__`` is mandatory in django>=1.9.

### DIFF
--- a/filer/fields/folder.py
+++ b/filer/fields/folder.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 
 import warnings
 
+from django import VERSION as DJANGO_VERSION
 from django import forms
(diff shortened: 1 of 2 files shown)


filer/fields/folder.py
```diff
@@ -122,6 +123,8 @@ def __init__(self, **kwargs):
                 )
                 warnings.warn(msg, SyntaxWarning)
         kwargs['to'] = dfl
+        if DJANGO_VERSION >= (1, 9):
```
jrief, line 126: current versions of django-filer do not support Django<1.11, so make a case distinction is unnecessary here.

PR #1203, opened 2020-08-20
Use dashed name for django-polymorphic in setup.py

ruohola:
The name with an underscore works, but just looks off in a `requirements.txt`. Also it will create unnecessary diffs when using tools such as [prequ](https://github.com/suutari/prequ) when an other `setup.py` has specified the name with a dash.


yakky (approved):

ruohola:
Could this be merged? 🧐

FinalAngel:
thank you @ruohola , could you also add a changelog entry for this?

ruohola:
> thank you @ruohola , could you also add a changelog entry for this?

Done.

FinalAngel:
thank you @ruohola !

PR #1206, opened 2020-09-03
Bumped version to 2.0.0

FinalAngel:
* Added support for Django 3.1
* Dropped support for Python 2.7 and Python 3.4
* Dropped support for Django < 2.2
* Changed the preferred way to do model registration via model inheritance
  and ``mptt.AlreadyRegistered``, which is deprecated since django-mptt 0.4
* Use dashed name for django-polymorphic dependency in setup.py

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ CHANGELOG
 =========
 
 
-2.0.0 (unreleased)
+2.0.0 (2020-09-03)
 ==================
 
 * Added support for Django 3.1
--- a/filer/__init__.py
+++ b/filer/__init__.py
@@ -14,6 +14,6 @@ Release logic:
 10. twine upload dist/django-filer-{new version}.tar.gz
 """
 
-__version__ = '1.7.1'
+__version__ = '2.0.0'
 
 default_app_config = 'filer.apps.FilerConfig'


FinalAngel:
@yakky good from your end to release?

jrief:
If we increase to a new major release, please really consider to merge #1152 !
Having the MIME-type of a file stored within the `File` object is a really important prerequisite for all kind of new features, such as SVG support.

FinalAngel:
@jrief is it a breaking change for existing projects as well? /cc @yakky

yakky:
> @jrief is it a breaking change for existing projects as well? /cc @yakky

It only contains a minor breaking change for a narrow use case, and I think 2.0 it's a proper way to communicate this

I'm :100: on adding #1152 in 2.0. It's ready for merge and it only needs a minor edit in the documentation (I'm adding a comment for it)

jrief:
@FinalAngel @yakky 
>  is it a breaking change for existing projects as well?

Yes, therefore I wanted to wait for a major version jump before merging.

However, most projects won't be affected. Only those using a custom class inheriting from `BaseFile` have to adjust the signature of `BaseFile.matches_file_type()`. This method now additionally takes the MIME-type.
 
It also could be, that some results change depending on the operating system. Reason is that until now, the MIME-type was detected using the ending of the file and a lookup table, which is quite unportable and had to be done over and over again.

With the #1152 change, the original MIME-type as sent by the browser during upload, is used.

FinalAngel:
let's get it in then :)

FinalAngel:
lets get it in then 👍

PR #1207, opened 2020-09-03
Get develop into master

FinalAngel:
* In ``models.File``, add field ``mime_type`` to store the Content-Type as set by
  the browser during file upload.
* For extended Django Filer models, adopt the classmethod ``matches_file_type`` to its
  new signature.
* Add attribute ``download`` to the download link in order to offer the file
  under its original name.


FinalAngel:
@jrief and @yakky here to PR to get it into develop

FinalAngel:
After this is merged, I can release - any other PRs to merge as well?

yakky:
@jrief is much more up to date than me regarding important PRs to be merged

jrief:
> After this is merged, I can release - any other PRs to merge as well?

I just went through all open pull requests. The only one, I would consider to merge is #958. This however has to be slightly changed (requested by @yakky, but never done by the original author).

Otherwise I'm OK with merging this pull request.

Just one side-note: I would really appreciate, if we could add a JSON field to the `File` class in the future. With Django-3.1 this wouldn't even required any third party dependency. Such a field would allow us to store arbitrary data, such as EXIF-headers, and other useful stuff without having to override the class. – just my two cents

FinalAngel:
@jrief that's probably best to discuss with @divio/django-cms-core

PR #1215, opened 2020-09-11
Bypassing the thumbnailing step for SVG file  (on master branch)

haricot:
related #460

This PR allows we to import svg files as Filer.Image model.
In the directory_table.hml file, the SVG file can be viewed if its size does not exceed 20 KB. Otherwise, in other cases view, the SVG files are displayed in the filer.


jrief:
Many thanks @haricot for this pull request. Now that we can handle SVG images through easy-thumbnails in a much more generic way, do you think this pull request adds features not provided by https://github.com/django-cms/django-filer/tree/svg-support ?

haricot:
turning the image into a canvas is misinterpreted:
use svglib reporlab only for the bound and modify the files with regex.

canvas loose info:this does not generate the gradient for example.
canvas_loose_info

haricot:
I think there shouldn't be a preview in the folder that lists the files in case there is a file imported that is too big, which can be negligible.
https://github.com/django-cms/django-filer/blob/cf6e35ddf2d8ab162b09a0e94323f4f173ed3f6f/filer/templates/admin/filer/folder/directory_table.html#L106-L110

Otherwise your approach is much better, this adds the computation of the crop and mine_type. So I can close this pull request which is not as deep.

PR #1224, opened 2020-11-03
Upgrade to work with NodeJS version 14

jrief:
It appears that the `gulpfile.js` has been written for a version of NodeJS, which meanwhile is out of date. Currently NodeJS 14.15.0 LTS is recommended for most users and presumably many of us have it installed.

In order to be able to compile SCSS and JavaScript, we have to adopt the build scripts. This pull request does that.


tests/frontend/unit/test.focal-point.js
```diff
@@ -61,7 +61,7 @@ describe('Cl.FocalPoint', function () {
             });
 
             expect(updateLocationValueStub).toHaveBeenCalled();
-            expect(updateLocationValueStub.calls.count()).toBe(1);
+            expect(updateLocationValueStub.calls.count()).toBe(2);
```
vxsx, line 64: Not really sure why would this change, since it doesn't seem to be failing on master, but the rest looks good!

vxsx (approved):

tests/frontend/unit/test.focal-point.js
jrief: apparently the testing framework now counts the invocations globally.

PR #1226, opened 2020-11-24
SVG support for django-filer

jrief:
This pull request adds a feature many users are waiting for: Uploaded images can now also be in SVG.
The heavy lifting of cropping and scaling SVG images is done inside easy-thumbnails (you need version 2.8rc0 or later).

Changes:
* Drop support for high resolution images and remove `retina.js` from project. High resolution images now are handled by the attribute in `` – as the HTML standard requires.
* `_width` and `_height` in model `Image` changed to `FloatFields`, this because SVG images specify their size as floats.
* For local development switched to NodeJS version 14. See also PR #1224 
* Add templatetag `file_icon` to `file_admin_tags`. It handles the rendering of all file types. 
* Remove pre-thumbnailing of images. Up to now, after upload images were scaled into many different sizes, most of which never were used. Now thumbnailing for rendering images in the admin is lazy.
* All icons for displaying folders, files (not images) have been replaced by nicer looking SVG variants.
* Increase size of thumbnails in list view from 25x25 to 40x40 pixels. 
* Added a playing facility for uploaded audio and video files.
* Fix scaling of very wide but short images – causing a division by zero exception: ceil height to integer.
* Add method `exists()` to `MultiStorageFieldFile`, which checks if a file exists on disk.
* Drop support of Python-3.5. Reason: Third party requirement `reportlabs` requires Python>=3.6.

I would propose version number 2.2 for the upcoming release containing these changes.


filer/models/abstract.py
```diff
@@ -84,7 +89,7 @@ def _get_exif(self):
         if hasattr(self, '_exif_cache'):
             return self._exif_cache
         else:
-            if self.file:
+            if self.file.exists() and self.mime_type != 'image/svg+xml':
```
yakky, line 92: Maybe we can just rewrite this by handling an exception in case `get_exif_for_file` fails? IDK if it can fail for other reason than file just being an svg

jrief: There is an even nicer solution: `cached_property`.

PR #1248, opened 2021-06-21
fixes #1247: see unclassified files when perms are on

### LINKED ISSUE #1247: not owned files in unfiled folder can not be listed if perms ON
If there are files in "unfiled folder" which are not owned by current user, he can not list those in admin because current code filters on files inside folders the user can see (but "unfiled folder" is virtual). 

Proposal: replace this line (filer/admin/folderadmin.py):

```
file_qs = file_qs.filter(models.Q(folder__id__in=perms) | models.Q(owner=request.user))
```

by those ones:
```
filters = models.Q(folder__id__in=perms) | models.Q(owner=request.user)
if isinstance(folder, UnsortedImages):
    filters = filters | models.Q(folder_id__isnull=True)
file_qs = file_qs.filter(filters)
```

**EDIT**
Bug still there when searching a file inside all files. So, replacement should be :

```

            file_qs = file_qs.filter(models.Q(folder__id__in=perms) 
                                     | models.Q(folder_id__isnull=True) 
                                     | models.Q(owner=request.user))

```

DylannCordel:
cf #1247

### DIFF
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -322,7 +322,11 @@ class FolderAdmin(PrimitivePermissionAwareModelAdmin):
         perms = FolderPermission.objects.get_read_id_list(request.user)
         root_exclude_kw = {'parent__isnull': False, 'parent__id__in': perms}
         if perms != 'All':
-            file_qs = file_qs.filter(models.Q(folder__id__in=perms) | models.Q(owner=request.user))
+            file_qs = file_qs.filter(
+                models.Q(folder__id__in=perms)
+                | models.Q(folder_id__isnull=True)
+                | models.Q(owner=request.user)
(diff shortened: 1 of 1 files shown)


jrief:
Thanks for patching.

Probably it would even be faster and cleaner, if we'd use the [Subquery](https://docs.djangoproject.com/en/2.2/ref/models/expressions/#subquery-expressions) expression, as introduced in Django-1.11.

What do you think?

jrief (approved):

DylannCordel:
I'm not sure about how you want to use subquery ? Would it be for the `perms` ? Because `perms` is either a `str` or a `set`, not a query (neither a `list` as the method's name (`get_read_id_list`) might lead us to think).

PR #1252, opened 2021-07-19
Remove legacy css

jrief:


jrief:
I had to change the code of django-filer, after configuring `django.contrib.staticfiles.storage.ManifestStaticFilesStorage` instead of the default. Some legacy CSS code references files, which no longer are part of the project. Therefore I removed that code.

PR #1308, opened 2022-08-12
Fix: Flie.objects.only("id") query when deleting a user.

wiltso:
## Description

When deleting a User object, Django's `django.db.models.deletion.Collector` will make a query like this: File.objects.filter(owner=user).only("id").
This then leads to a recursion error due to that the `filler.models.File` object requires that the following fields are present:
- _file_size
- sha1
- is_public

## Related resources

- https://github.com/django-cms/django-filer/issues/1290

## Checklist


vinitkumar:
@wiltso Do you remember if the error you got is almost like this?

```python
IntegrityError at /admin/customroles/siteuser/
update or delete on table "auth_user" violates foreign key constraint "filer_file_owner_id_b9e32671_fk_auth_user_id" on table "filer_file"
DETAIL: Key (id)=(89) is still referenced from table "filer_file".
```

wiltso:
I was actually getting an recursion error.
You can replicate it by doing `File.objects.filter(owner=u).only("id")`.

fsbraun:
@wiltso Sorry, that this got "lost". I would like to take it up. I'd really appreciate, if we could add a test for it:

* Create user
* Create filer object
* Delete user

Do you think you could add such an test?

fsbraun:
@wiltso I've created a follow-up PR (#1357) that includes the cherry-pick of this PR and a simple test. 
I really would like to get this fix merged now.

fsbraun:
Implemented in #1357

Issue #1317, opened 2022-10-02
Choose file popup doesn't select image, it exits with no image selected


ivandir:
Choose file popup doesn't select image with Django 4.1.1. I kept downgrading based on django-filer dependency and the following worked.

Downgrading to 2.2.2 selected Django 4.0.7 - this worked
Upgrading to 2.2.3 and forcing Django 4.0.7 - this worked and one I'm currently using

Don't upgrade to to 2.2.3 and forcing Django 4.1.1 - this is broken

rezamazarloo:
i have the same issue.

rezamazarloo:
in my case the problem does not solve by downgrading.
i think the problem is somewhere else.
i can upload image with drag and drop but the choose file doesn't select any image when i select an image, nothing happens and it seems the image is not selected in django admin.

linhvng:
I'm using Django 4.1.3 and have the same issue. I can upload by drag and drop the image but cannot choose already uploaded files.

fsbraun:
Django filer does not yet support Dango 4.1. can you test if the issue remains if you use this version: `pip install git+https::/GitHub.com/fsbraun/django-filer@feat/django-4-1`?

fsbraun:
@linhvng @rezamazarloo @ivandir Could you check if the latest master branch solves your issue? You can install it using `pip install git+https://github.com/django-cms/django-filer@master#egg=djangocms-filer`. It you be great if #1326 solved your issue!

fsbraun:
@marksweb Should we put this into a minor release?

marksweb:
@fsbraun yeah that sounds sensible to me

PR #1326, opened 2022-12-20
Fix popup window name to element id transform

stefanw:
## Description

Django 4.1 adds a [popup index to the admin popup window name](https://github.com/django/django/commit/492ed60f236d770eb9a6d56d85ff2550bb1ecfff) that needs to be stripped when getting back to the element id.

This should be backwards compatible to previous element ids unless they use double underscores (`__`) in them.

## Checklist


fsbraun:
Could that be the root cause of #1317 ?

stefanw:
Yes, sounds very much like what I experienced.

fsbraun (approved):

fsbraun:
Well spotted!! Thank you!

PR #1344, opened 2023-05-10
Modernize django filer

jrief:
## Description

superseeds #1341 

* Remove broken PR #1255 
* Replace re_path against path
* Use admin.decorators.action to mark actions
* Remove dysfunctional jQuery function $.actions()

## Checklist

* [x] I have opened this pull request against ``master``
* [ ] I have added or modified the tests when changing logic
* [x] I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [x] I have read the [contribution guidelines ](https://github.com/django-cms/django-filer/blob/master/docs/development.rst#contributing) and I have joined #workgroup-pr-review on 
[Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

Issue #1343, opened 2023-05-10
Thumbnail view does not work as expected


jrief:
That new thumbnail view really looks great, but unfortunately it does not work properly.
@corentinbettiol please check this, because it's a show-stopper for that feature.

## How to reproduce
* Use django-filer version 2.3rc1 or HEAD in branch master.
* Open a directory list view, for instance http://my-site/en/admin/filer/folder/, it should contain at least 1 file or folder.
* Click on the checkbox on the left side of that folder.
* In the toolbar there are buttons to delete, copy and move the selected files. In an additional drop-down menu there are options to change permissions, to rename and to resize the selected images. If at least one file is selected, it is possible to perform these actions.
* Switch over to thumbnail view and try the same. Even if a file is selected, those action buttons are disabled.

Screenshot:
<img width="178" alt="Screenshot 2023-05-10 at 16 24 07" src="https://github.com/django-cms/django-filer/assets/1132472/0841003c-9262-4bc2-b540-d1e5f8f2382b">

fsbraun:
I think there's a solution: #1345

PR #1255, opened 2021-09-09
Add thumbnail view for folder list admin view (updated)

corentinbettiol:
Hi :)

Since it's way more easier to manage images in django-filer with bigger thumbnails instead of the small ones we have in the list, I wanted to add a thumbnail view.

This PR aims to solve https://github.com/django-cms/django-filer/issues/723.

I saw that the PR https://github.com/django-cms/django-filer/pull/1076 already included all the required changes, but for an unknown reason wasn't merged back in 2018. So I cherry-picked the commits and backported them onto the master branch on our fork :)

[Here's a demo video](https://share.kapt.mobi/files/1631203999-django-filer-demo.webm) (feature is shown at ~ 15 seconds).

PR #1345, opened 2023-05-12
fix: Broken action buttons in thumbnail view 

fsbraun:
## Description

I propose this PR as an alternative to removing #1255 in #1344. **It fixes the dysfunctional UI without removing functionality from the repo**:

* The `$.actions()` extension of `django.jQuery` is not present in current versions of Django. This PR removes them. The call of `$.actions()` lead to the JS execution to be stopped only in the thumbnail view. In the table view it just raised an error on the console.
* Django's admin JS sets the `selected` class automatically in the table view, but not in the thumbnail view. This has been added and leads to the activation of the action buttons in the thumbnail view.
* For no apparent reason, the thumbnail view's JS waited for the page to be loaded twice.
* Also updates the jQuery UI version

No changes are made to the python code base (and the improvements to it from #1344 would need to be added in a second PR).

## Related resources

* #1344 
* #1343 
* #1255
* [Django admin's legacy actions jQuery extension](https://github.com/django/django/blob/stable/3.1.x/django/contrib/admin/static/admin/js/actions.js) which was removed in Django 3.2

## Checklist


marksweb (approved):
Thanks Fabian! 🎉

PR #1353, opened 2023-06-10
feature: Significant performance boost for filer directory listings

fsbraun:
## Description

When removing `mptt` #1351 once concern were database hits. While getting the full path of a folder with and without `mptt` requires `n` db hits, `n` being the nesting of the current folder typically is small. 

For directories containing `m` files, the performance bottleneck appeared to be getting the thumbnails requiring `m` database hits. Filer limits the number of files per page to 20 to avoid this problem to take excessive time.

This PR builds upon #1351 and improves directory listings in filer by
* **Annotating the expected paths of a file's thumbnail URLs** (regular, and double resolution). That's a bit tricky and requires Django subqueries. It gives a great **heuristic** for checking if a relevant thumbnail already exists. If so, it is used without having easy thumbnails, hit the db again to check.
* **Avoid creating missing thumbnails during the directory request**. Instead, an additional admin URL generates missing thumbnails in a separate request each (potentially taking advantage of multiple worker instances). 
* **Use `select_related` on a folder's and file's `owner`.** This was a classical `m+1` issue with the `User` object.
* **Cache template tag `file_icon_url`** which is called several times per file when rendering a directory. Also, it needlessly checked for the existence of thumbnails in the db.
* For getting thumbnails, **do not check if the original file exists**, if not on the `FileSystemStorage`: Otherwise this tests triggers network connections to the file storage (such as AWS) for each file in the directory.

Accompanying changes are:
* I have set the default **pagination limit from 20 to 100**. 
* **Thumbnails are loaded lazily** to balance the load a little.

Further performance improvement lever:
* Configure and use easy-thumbnail's celery integration to have necessary thumbnails always at hand

## Performance impact

This is just one example: 71 files at nesting level 2 requires 

DB hits when ...    | Current filer | This PR
------------------| ------------| --------
Listing directory   | 301 | 13
Selecting file        | 952 | 13

DB queries are **independent of folder size** (in terms of number of files/folders in that folder) as opposed to between `4m` and `13m`.

## Way forward

This performance improvement is so significant that I suggest dropping the `mptt` dependency and merging the request for version 3.0 even before any major redesign for filer (see #1348). (In a separate upcoming PR, I would like to address minor styling issues, which might also go into 2.3).

## Related resources

* #1351 
* #...

## Checklist


tests/test_admin.py
```diff
@@ -90,6 +90,27 @@ def test_filer_directory_listing_root_get(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['folder'].children.count(), 6)
 
+    def test_filer_directory_listing_performance(self):
+        images = 10
```
vinitkumar, line 94: Let's increase this number to some 100 images? That would really test the perf I think?

vinitkumar (approved):
Great work here @fsbraun. It also helps me with an issue at work. You have my 👍🏼

tests/test_admin.py
fsbraun: Hi @vinitkumar ! Any number > 7 (no of queries expected) will ensure the there is no query for each file. i took 10 to save testing time and resources.

PR #1359, opened 2023-06-15
Feat/filer perm improvements

### LINKED ISSUE #1358: Issues with permissions in Django filer
1. Issue with adding folder permission when selecting all files crashing due to gettext_lazy
2. Perms should checked for non-root users in the unsorted upload to verify if they should access the file
3. Add additional check in template that you are only able to see your own files and also the change buttons as well.

vinitkumar:
## Description

Implement some permissions related fixes for the filer.

## Related resources

Refer to issue, resolves #1358 

## Checklist

### DIFF
--- a/filer/models/virtualitems.py
+++ b/filer/models/virtualitems.py
@@ -48,8 +48,20 @@ class UnsortedImages(DummyFolder):
     is_unsorted_uploads = True
     _icon = "unfiled_folder"
 
+    def __init__(self, user=None):
+        super().__init__()
+        self.user = user
+
     def _files(self):
-        return File.objects.filter(folder__isnull=True)
+        """
+        If the current user is not a superuser, we don't them to see/edit/change
+        file in the unsorted uploads folder that they don't own.
+        """
+        if self.user.is_superuser:
+            return File.objects.filter(folder__isnull=True)
+        else:
+            return File.objects.filter(folder__isnull=True, owner=self.user)
+
     files = property(_files)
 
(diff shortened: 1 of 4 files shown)


filer/models/virtualitems.py
```diff
@@ -48,8 +48,20 @@ class UnsortedImages(DummyFolder):
...
+        super().__init__()
+        self.user = user
+
     def _files(self):
-        return File.objects.filter(folder__isnull=True)
+        """
+        If the current user is not a superuser, we don't them to see/edit/change
+        file in the unsorted uploads folder that they don't own.
+        """
+        if self.user.is_superuser:
```
fsbraun, line 60: For backwards compatibility, you could change this to
```suggestion
        if self.user.is_superuser or not filer_settings.FILER_ENABLE_PERMISSIONS:
```
This should still show all files if `FILER_ENABLE_PERMISSIONS` is `False`

filer/models/virtualitems.py
```diff
@@ -48,8 +48,20 @@ class UnsortedImages(DummyFolder):
...
+            return File.objects.filter(folder__isnull=True)
+        else:
+            return File.objects.filter(folder__isnull=True, owner=self.user)
```
fsbraun, line 63: A test would be great... ;-)

fsbraun (commented):
Great work! I just have two comments:
* Can we add a test to see if other peoples' files are really hidden?
* I'd still show all files in the general folder if permissions are not switched on.

Issue #1392, opened 2023-07-19
Copy folder is broken after mptt removal


protoroto:
After django-mptt removal, https://github.com/django-cms/django-filer/blob/master/filer/admin/folderadmin.py#L1083 this will fail as `insert_at` is a django-mptt method.

I'll open a pr right now with two commits, one with a regression test (folder copying was never tested) and one with the fix.

I'm not sure if it is the right way of doing things, please tell me if something is wrong.

PR #1393, opened 2023-07-19
Fix copy folder being broken after django-mptt removal

protoroto:
## Description

After `django-mptt` removal, copying folders is broken because `insert_at` method is not there anymore.
I've added a regression test that should show the error, and then a fix for it that make the test pass.

## Related resources

* #1392 

## Checklist

### DIFF
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -1076,7 +1076,7 @@ class FolderAdmin(PrimitivePermissionAwareModelAdmin):
 
         old_folder = Folder.objects.get(pk=folder.pk)
 
-        new_folder = Folder.objects.create(
+        folder, _ = Folder.objects.get_or_create(
             name=foldername,
             owner=old_folder.owner,
             parent=destination,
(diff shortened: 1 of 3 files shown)


fsbraun (approved):
Thank you! Well spotted and excellent work!

filer/admin/folderadmin.py
```diff
@@ -1076,19 +1076,19 @@ def _copy_folder(self, folder, destination, suffix, overwrite):
 
         old_folder = Folder.objects.get(pk=folder.pk)
 
-        # Due to how inheritance works, we have to set both pk and id to None
-        folder.pk = None
-        folder.id = None
-        folder.name = foldername
-        folder.insert_at(destination, 'last-child', True)  # We save folder here
+        new_folder = Folder.objects.create(
```
vinitkumar, line 1079: I think doing 'get_or_create' is getter to prevent exceptions. What do you think?

vinitkumar (changes requested):

filer/admin/folderadmin.py
vinitkumar: We can also just keep using folder as the variable name, it doesn't really make any difference other than the semantics.

vinitkumar: @fsbraun Do you agree?

protoroto: @vinitkumar @fsbraun Changed with `get_or_create`, it does look better.

vinitkumar (approved):

Issue #1394, opened 2023-07-19
'File' object has no attribute 'width'


wagnerfelix:
Hi,
we use our own Image model and we recently updated our code to work with 3.0.2 from 2.2.4. After upgrading we noticed that we could not access the details of files. The cause is  [fix: Prevent server errors in image detail view if images do not exist](https://github.com/django-cms/django-filer/pull/1389). Specifically `templates/admin/filer/tools/detail_info.html`, in line 6 the templatetag `{% file_icon original detail=True %} ` causes an issue. The "offending" code is in `filer/templatetags/filer_admin_tags.py`:

``` 
    ...   
    if detail:
        context['download_url'] = file.url
        if file.width:
            width, height = 210, ceil(210 / file.width * file.height)
            context['sidebar_image_ratio'] = file.width / 210
    ...
```
`'File' object has no attribute 'width'`. After downgrading to 3.0.1 the issue goes away.
Regards

PR #1395, opened 2023-07-19
fix: crash in the file detail view

vinitkumar:
## Description

In a recent fix, we introduced a regression where we started to access width and height of a file. However, this will only work for images and not for normal files like PDF, Docx etc.

In order to fix this, we need to guard and check if the file is actually an image before trying to access the attributes.

## Related resources

- Github Issue: #1394

## Checklist


marksweb:
Could you add a test or two to ensure this doesn't regress again?

vinitkumar:
> Could you add a test or two to ensure this doesn't regress again?

Done, please verify if it's correct.

PR #1416, opened 2023-09-01
Re-add alphabetical sorting as default (fixes #1415)

### LINKED ISSUE #1415: Sorting of Files seems broken for 3.0.x versions
Hey there,

we recently upgraded one of projects to django-filer 3.0.5 and it seems that the sorting logic is doing something funny.

In our old versions everything was sorted correctly, either by name of filename, and was even merged together (alas, when you had some files with a dedicated name and some without, therefore inheriting the label as their filename, everything was correctly sorted still.)

We already skimmed through the code to find some answers but couldn't really wrap our heads around the issue since nothing of importance seems to have been changed..

Is anyone else experiencing the same problems? Is it intentional? Or maybe even some kind of bug?

In either way, we're more than happy to support with code changes, just let us know if this is something which is, indeed, a bug and if not, how we could work around this new ordering mechanism.

Thanks and have a good day.

filipweidemann:
## Description

This PR intends to hotfix #1415 by reintroducing the `.sort()` function on `file_qs` inside the `filer/admin/folderadmin.py` logic. 

We did not modify any existing tests or added new ones because the function is returning rendered HTML and it is currently unclear how we could test this in an elegant way.

We are also aware that there are better ways to accomplish the same result, e.g. by using Django's `QuerySet.annotate` and `Coalesce` DB functions to add a correct sort target to the queryset, so that `.order_by()` works as intended.

However, we did not want to refactor the existing solution right now but rather hotfix the (currently broken) behaviour.

## Related resources

* See issue #1415 
* https://docs.djangoproject.com/en/4.2/topics/db/aggregation/
* https://docs.djangoproject.com/en/4.2/ref/models/database-functions/#coalesce

## Checklist

### DIFF
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -317,13 +317,14 @@ class FolderAdmin(PrimitivePermissionAwareModelAdmin):
 
         folder_qs = folder_qs.order_by('name').select_related("owner")
         order_by = request.GET.get('order_by', None)
-        if order_by is None:
-            order_by = "file"
-        order_by = order_by.split(',')
-        order_by = [field for field in order_by
-                    if re.sub(r'^-', '', field) in self.order_by_file_fields]
-        if len(order_by) > 0:
-            file_qs = file_qs.order_by(*order_by)
+        if order_by is not None:
+            order_by = order_by.split(',')
+            order_by = [field for field in order_by
+                        if re.sub(r'^-', '', field) in self.order_by_file_fields]
+            if len(order_by) > 0:
+                file_qs = file_qs.order_by(*order_by)
+        else:
+            file_qs.sort()
 
         if folder.is_root and not search_mode:
             virtual_items = folder.virtual_folders
--- a/filer/admin/folderadmin.py
+++ b/filer/admin/folderadmin.py
@@ -380,7 +380,7 @@ class FolderAdmin(PrimitivePermissionAwareModelAdmin):
             }
         except:  # noqa
             permissions = {}
-        
+
         items = list(itertools.chain(folder_qs, file_qs))
         paginator = Paginator(items, FILER_PAGINATE_BY)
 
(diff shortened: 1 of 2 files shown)


fsbraun:
@filipweidemann Thanks for creating this PR! Just a quick question: Since you seem to have run `black` it is hard for me to identify the actual content changes. I assume it's the `file_list`, its sorting and passing the `file_list` into the context if it exists. Did I miss anything?

Can you add a test that verifies files are sorted? This feature got lost since a test did not cover it. It would be great if we fixed this for the future.

filipweidemann:
@fsbraun okay good catch, running `black` was not intentional, I just use it for everything by default and forgot to turn it off. I'll fix that and try to add some tests as well.

filipweidemann:
Okay  @fsbraun, here we go.

I reverted the `black` formatting and added some tests with the expected sort behaviour.

I really was not happy with using `QuerySet` objects by default and falling back to a `list` just to do the sorting inside the Python code.

So I did a quick refactoring and implemented the solution I already talked about inside the PR description.
Currently we are using `Coalesce` to create a temporary `coalesce_sort_field` that priorititzes the `name` field and falls back to the `original_filename` for all objects without a set `name`. I used `Case` to catch the empty strings of the `name` field since the default behaviour of `Coalesce` does not interpret empty strings and `NoneType` equally, but we want to fall back on empty strings as well.

Would be nice if you could have a quick look & maybe kick off the CI again.

If something is still unclear or needs another refactor, let me know. 
Thanks.

fsbraun (approved):
LGTM! Nice work!

filipweidemann:
Hey, quick question: when will this get merged & released? I don't want to pressure anyone but a quick estimate would be nice because right now I'm thinking about shipping the fork to prod environments to bridge the gap until the release is there.

But I would obviously avoid that if you're saying that this will be released this week or something :)

Thanks!

fsbraun:
@filipweidemann Will release it the next days as 3.0.6. (Merged right now. - I typically do not merge immediately after reviewing to give other community members a chance of looking at it.)

filipweidemann:
Great stuff, that's all I needed to know. :)
Thanks for explaining the workflow and handling the release!

Issue #1417, opened 2023-09-05
Double .open() causing issues with django-storages 1.14 and S3 backend


GaretJax:
django-storages 1.14 has been released with a bugfix for the `closed` File property: https://github.com/jschneier/django-storages/pull/1249

This causes the code at https://github.com/django-cms/django-filer/blob/master/filer/models/filemodels.py#L257-L260 to raise an [exception](https://github.com/django/django/blob/main/django/core/files/base.py#L114) and thus preventing `_copy_file` from being called.

At this point I'm unsure about what the fix would look like and whether that code is still required. If yes, it might be enough to check for `src_file.closed` before calling the second `.open()`.

fsbraun:
@GaretJax can you share the kind of issues you encountered?

fsbraun:
Fixed in 3.0.6

GaretJax:
Thanks @fsbraun; I somehow did not receive any notification from github for this activity.

fsbraun:
Pleasure. Good catch!!

PR #1418, opened 2023-09-06
fix: django-storage 1.14 complains about files being opened twice when copying

fsbraun:
## Description

While the root cause of the bug #1417 might be in django-storages 1.14, this quick-fix ensures that any ValueError raised opening a file for copying does not crash the file application with a server error.

## Related resources

* #1417 
* #...

## Checklist


vinitkumar:
@fsbraun Is it possible to have a test with this patch?

fsbraun:
@vinitkumar I know! I've been thinking about it, but I do not see a way of testing this without the specific implementation of `django-storage`'s S3Storage on an AWS instance. I frankly believe that the underlying issue is between Django and django-storage, but it's hard to tell. Since the`try` `except` block. 

Any idea how to test this?

vinitkumar:
@fsbraun I don't know. Seems hard to implement in the test. Have you tried testing this patch with a real s3 backend?

fsbraun:
@vinitkumar I tested it on my staging setup with AWS: Without the fix, copying files fails and a copy of the first file remains in the source folder. After fixing it, copying works fine.

The fix also does not affect local file storage (which just runs `seek(0)` in case of a reopen).

vinitkumar (approved):
Looks good to me. Let's get this merged and released 🚢  🚀

PR #1420, opened 2023-09-10
fix: import_files command to skip existing files

misli:
- Fix use of get_or_create to really skip existing files

## Description

The `import_files` command tries to skip existing files, but it always create new one,
because the way it detects the existing one is wrong.

## Checklist


fsbraun:
Hi @misli ! Thank you for the pull request. 

Can you elaborate on the issue? Why are existing files detected wrongly?

From the code I see, you allow to have multiple files with the same name in the same folder but then ignore all of them but the first. To me it is not immediately transparent why this should be the right way to go. But then again, I have not fully understood the issue.

PR #1426, opened 2023-09-24
fix: decompression bomb attack in the Filer

### LINKED ISSUE #1330: Restrict maximum image dimensions/resolution
With many server instances memory limited, I believe it would be useful to allow for maximum image dimensions to be checked and files rejected if larger.

As the memory usage is not internally checked, it is possible to exhaust the memory available and crash/slow down significantly any other request processing with just a single upload. We have found 6000x6000 pixels to be a safe limit for our case, though this would vary depending on the server resources available.

We have implemented the following monkey-patch to achieve this functionality:

`settings.py`

```
FILER_IMAGE_MAX_WIDTH = 6000
FILER_IMAGE_MAX_HEIGHT = 6000
```

Then we placed a monkey patch into one of our project files:

```
# Monkey-patch filer 2.2.4
import filer
from filer.models import Image
if filer.__version__ != "2.2.4":
    print("ERROR: Expecting filer version 2.2.4")
else: 
    print("Monkey-patching filer Image clean")
    prev_clean = Image.clean

    def clean(self):
        ret = prev_clean(self) # original            
        try:
            imgfile = self.file.file
        except ValueError:
            imgfile = self.file_ptr.file
        imgfile.seek(0)
        if self.mime_type == 'image/svg+xml':
            pass # noop, SVG image
        else:
            from filer.utils.compatibility import PILImage
            from sentry_sdk import capture_exception
            from django.core.exceptions import ValidationError     

            try:
                width, height = PILImage.open(imgfile).size
                print(width, height)
                if width > settings.FILER_IMAGE_MAX_WIDTH:
                    print("Check failed")
                    raise ValidationError({"file": "Image is too wide. Please refresh the tab to continue."})
                if height > settings.FILER_IMAGE_MAX_HEIGHT:
                    print("Check failed")
                    raise ValidationError({"file": "Image is too tall. Please refresh the tab to continue."})
            except ValidationError as ve:
                raise ve
            except Exception as e:
                print(e)            
                capture_exception(e)
                #raise Exception("Unable to process image: {}".format(e))
                raise ValidationError({
                    "file": "Unspecified error {}".format(e)
                })
        imgfile.seek(0)

        return ret
    
    Image.clean = clean # re-assign
```

The above seems to be working fine in the "full" admin view, however, no error is shown when large file is dragged into the file browser popup, nor there is any indication when dragged into the form field directly (the field just re-sets).

To start with though, I believe it would make sense to implement the size check as an option for Filer. Thoughts?

### LINKED ISSUE #1425: Issue with DecompressionBombWarning with high pixel images
Filer allows for images to be uploaded to the storages. There is a setting where we can limit the size of the image. However, there is no way to prevent `DecompressionBombWarning` that can happen when you have added a high pixel image which Pillow can't handle.

In that case you end up with errors like this

```python
/usr/local/lib/python3.9/site-packages/PIL/Image.py:3074: DecompressionBombWarning: Image size (94570638 pixels) exceeds limit of 89478485 pixels, could be decompression bomb DOS attack.
```

I believe, we should make use of the check for max allowed Pixel from Pillow and give a validation error before uploading such images as I have experience downtime in website because filer allowed such high pixel images to be uploaded and then crash the page as pillow can't handle it. Even the filers own thumb-nailing won't succeed because of this issue as it can' handle such files. I believe it is a real issue and filer is the correct place to solve this.

vinitkumar:
## Description

Filer allowed for the images for very high pixels (height * width) to be uploaded. This would cause crash and failures when the high pixels exceeded what is allowed by Pillow Image MAX_IMAGE_PIXELS value.

This is an issue because even though the image is possible to be created and attached to the page, it would never work as PIL always fails to thumbnails such high pixel image and crashes causing crash and high memory usages in such pages.

This patch, fixes this issues in the bud as it wouldn't allow such files to be uploaded via FILER itself. It also allows to set a lower limit FILER_MAX_IMAGE_PIXELS so that users can limit the max pixels to value much lower than what PIL support.

We also choose the pixel value rather than MAX_HEIGHT and MAX_WIDTH to allow different resolutions of image other than square images.

https://github.com/django-cms/django-filer/assets/537678/852f1ec2-ce5b-4eeb-80d1-f8eb070e5bfa

Github Issue 

- fixes #1425 
- fixes #1330

Authored-by: Vinit Kumar <vinit.kumar@kidskonnect.nl>

## Related resources

* https://en.wikipedia.org/wiki/Zip_bomb
* https://github.com/python-pillow/Pillow/blob/main/src/PIL/Image.py#L3148

## Checklist


filer/admin/clipboardadmin.py
```diff
@@ -124,8 +136,35 @@ def ajax_upload(request, folder_id=None):
...
-        file_obj.save()
+        if isinstance(file_obj, Image):
+            # We check the Image size and calculate the pixel before
+            # the image gets attached to a folder and saved. We also
+            # send the error msg in the JSON and also post the message
+            # so that they know what is wrong with the image they uploaded
+            from django.contrib.messages import ERROR, add_message
+            pixels: int = max(1, file_obj.height) * max(1, file_obj.width)
+
+            if pixels > 2 * FILER_MAX_IMAGE_PIXELS:
```
fsbraun, line 147: Quick question: Why `2 * FILER_MAX_IMAGE_PIXELS`?

filer/admin/clipboardadmin.py
```diff
@@ -124,8 +136,35 @@ def ajax_upload(request, folder_id=None):
...
+                msg = (
+                    f"Image size ({pixels} pixels) exceeds limit of {2 * FILER_MAX_IMAGE_PIXELS} "
```
fsbraun, line 149: We should make this string localizable.

filer/admin/clipboardadmin.py
```diff
@@ -124,8 +136,35 @@ def ajax_upload(request, folder_id=None):
...
+                    "pixels, could be decompression bomb DOS attack."
+                )
+
+                message = str(msg)
+                add_message(request, ERROR, message)
+                return JsonResponse({'error': message})
+
+            if pixels > FILER_MAX_IMAGE_PIXELS:
+                msg = f"Image size ({pixels} pixels) exceeds limit of {FILER_MAX_IMAGE_PIXELS} pixels, could be decompression bomb DOS attack."
```
fsbraun, line 158: Also needs to be localizable.

Also, maybe (just a though) give an example image size that works :
```
aspect = file_obj.width / file_obj.height
res_x = int((FILER_MAX_IMAGE_PIXELS * aspect)^0.5)
res_y = int(res_x / aspect)
```

Now, `res_x` and `res_y` could be used as maximum size with a given aspect ratio. This would allow the user to resize the image before uploading.

fsbraun (changes requested):
Great work and immediate UI response! This does also work when drag-and-dropping?

Can you add a test based on `test_filer_upload_file` which the `FILER_MAX_IMAGE_PIXELS` set to a very low value to see if the error messages are triggered?

filer/admin/clipboardadmin.py
vinitkumar: Just used the same logic as in here: 
https://github.com/python-pillow/Pillow/blob/main/src/PIL/Image.py#L3148

filer/admin/clipboardadmin.py
vinitkumar: Agree.

vinitkumar:
@fsbraun Just checked it out. Looks and work great. Except I think this warning message could be better. Is show `__all__` and some message.

Screenshot 2023-09-28 at 2 23 43 AM
O
Screenshot 2023-09-28 at 2 23 56 AM
Once that is fixed. I think it's good enough to be merged and we can cut a 3.1 release for filer.

fsbraun (approved):

Issue #1427, opened 2023-09-25
Directory table list item's URL action should not trigger a download


T4m:
In the CMS admin, when listing the files, we have up to 4 action buttons, with the two firsts "URL" & "Download".

Before this [change](https://github.com/django-cms/django-filer/pull/1360/files#diff-714fc9b6a87e4507da2245171743fa7245805d521413cbcfe23074dc80ebe623R143) : **Both were links** (one canonical, one to the file in media).

**Now both are download links**.

IMHO : 
* the "download" attribute on the "canonical url" action, is a mistake and should not be there.
* the "download" attribute on the "download" action, is a good and correct improvement to the code.

It became more difficult to display the file in the browser without downloading it. (Now only "right click > open url" allows it.)

So, in short, line 143 in this version of "filer/templates/admin/filer/folder/directory_table_list.html" should be removed.

Also, I noted that the feature name of the PR is "Styling update", but this change is really a behavior update.

T4m:
I created the related [pull request](https://github.com/django-cms/django-filer/pull/1428)

fsbraun:
@T4m There's a security concern allowing to open files directly in the browser window: SVG filers may contain potentially malevolent JavaScript code. Starting from filer 3.1 the plan is as follows:
* Show images using the expand button from the image details view ("change view"). SVGs are rendered using an `<img>` tag, other images are opened directly in the browser.
* Canonical URL action will now copy the canonical URL to the clipboard and not open the image.
* Right-clicking on a canonical URL action will allow opening the file in a new tab/window. SVGs are the n rendered using the `<img>` tag.

I hope this is an acceptable compromise between security and UX. @T4m What are your thoughts?

PR #1428, opened 2023-09-25
fix: URL action in data table list template

T4m:
## Description

In the CMS admin, when listing the files, we have up to 4 action buttons, with the two firsts "URL" & "Download".

Before this [change](https://github.com/django-cms/django-filer/pull/1360/files#diff-714fc9b6a87e4507da2245171743fa7245805d521413cbcfe23074dc80ebe623R143) : Both were links (one canonical, one to the file in media).

Now both are download links.

IMHO :
* the "download" attribute on the "canonical url" action, is a mistake and should not be added.
* the "download" attribute on the "download" action, is a good and correct improvement to the code.

It became more difficult to display the file in the browser without downloading it. (Now only "right click > open url" allows it.)

So this PR fixes this accordingly.

Also, I noted that the feature name adding the bug was "Styling update", but this change was more a behavior update, I think. So I didn't add anything to the changelog of my PR because it's more "going back to the documented state". However, it might be relevant to document somewhere that the previous version added the download attribute on the download link.

### DIFF
--- a/filer/templates/admin/filer/folder/directory_table_list.html
+++ b/filer/templates/admin/filer/folder/directory_table_list.html
@@ -140,7 +140,6 @@
                             <td class="column-action">
                                 {% if file.canonical_url %}
                                     <a href="{{ file.canonical_url }}" rel="noopener noreferrer"
-                                       download="{{ file.original_filename }}"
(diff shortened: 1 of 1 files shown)


fsbraun (approved):

fsbraun:
This solves #1427

PR #1431, opened 2023-10-02
fix: initialize dropzoneSelector inside inline formsets in a Django 4.1 compatible way

### LINKED ISSUE #1400: FilerImageField drag'n'drop not working on inlines
Using Django 4.1.9 with django-filer 2.2.5 but I have also tried with higher versions.

It happens only on adding new inlines (not on already visible).

I assume JS for dropzone is not initialized when adding new inlines.

mbi:
## Description

Django 4.1 changed the JavaScript event type being fired when adding a new formset from a jQuery to a vanilla JavaScript event. https://docs.djangoproject.com/en/4.1/ref/contrib/admin/javascript/

This broke the drag-and-drop feature when adding dynamic inlines containing Filer fields in e.g. the Django admin, because the dropzone failed to initialize.

This patch checks for Django 4.1 and initializes the dropzone on newly added inline rows.

## Related resources

This fixes #1400.
## Checklist

### DIFF
--- a/filer/static/filer/js/addons/dropzone.init.js
+++ b/filer/static/filer/js/addons/dropzone.init.js
@@ -164,7 +164,27 @@ djQuery(function ($) {
         // There is no way to feature detect the new behavior implemented in Django 1.9
         if (!window.__admin_utc_offset__) {
             $(document).on('formset:added', function (ev, row) {
-                var dropzones = $(row).find(dropzoneSelector);
+                if(ev.detail.formsetName) {
+                    /*
+                        Django 4.1 changed the event type being fired when adding
+                        a new formset from a jQuery to a vanilla JavaScript event.
+                        https://docs.djangoproject.com/en/4.1/ref/contrib/admin/javascript/
+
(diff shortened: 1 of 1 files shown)


fsbraun:
@mbi Is there any way to test this? How did you test?

mbi:
@fsbraun I tested manually, i.e. with an Django admin inline containing a FilerImageField. 

Without the patch, when you add a new row and drag a file onto the field, it doesn't react. With the PR applied, the field will react correctly and upload the file when dropped.

mbi:
@fsbraun I just tested this on a 3.2 project, it works fine (as it was before: the new code line only triggers on 4.1+)
I also remove the two blocks supporting Django < 1.9

marksweb (approved):
This sounds straightforward enough and it's good to see the explanation of what's happening.

@fsbraun I'm always in favour of dropping support for unsupported Django. If we maintain support we're effectively encouraging people to use unsupported versions.

fsbraun:
Thanks, @mbi ! Great work!

PR #1434, opened 2023-11-04
feat: Image dimensions update management command

benzkji:
## Description

Basic management command, to insert missing image dimensions.
I've added a test, but can't find instructions on how to setup local tests.

Also, I've only seen conventional commits just now...so...

### DIFF
--- a/filer/management/commands/filer_check.py
+++ b/filer/management/commands/filer_check.py
@@ -148,8 +148,10 @@ class Command(BaseCommand):
                 imgfile.seek(0)
             except (FileNotFoundError):
                 continue
-            if image.file.name.endswith('.svg'):
-                image._width, image._height = VILImage.load(imgfile).size
+            if image.file.name.lower().endswith('.svg'):
+                with VILImage.load(imgfile) as vil_image:
+                    # invalid svg doesnt throw errors
+                    image._width, image._height = vil_image.size
             else:
                 try:
                     with PILImage.open(imgfile) as pil_image:
--- a/filer/management/commands/filer_check.py
+++ b/filer/management/commands/filer_check.py
@@ -42,6 +42,13 @@ class Command(BaseCommand):
             help="Delete references in database if files are missing in media folder.",
         )
         parser.add_argument(
+            '--image-dimensions',
+            action='store_true',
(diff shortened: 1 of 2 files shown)


benzkji:
I guess I am doing too much work here. It would probably better to call some helper function of filer? Change a field, call save(), and thus triggering calculation of _width and _height (and possibly other meta fields).

Also, my test fails, probalby because the provided JPG is not a real one?

filer/management/commands/filer_check.py
```diff
@@ -112,3 +121,30 @@ def walk(prefix):
...
+            try:
+                imgfile = image.file.file
+            except ValueError:
+                imgfile = image.file_ptr.file
+            imgfile.seek(0)
+            if image.file.name.endswith('.svg'):
+                image._width, image._height = VILImage.load(imgfile).size
+            else:
+                pil_image = PILImage.open(imgfile)
+                self._width, self._height = pil_image.size
```
fsbraun, line 147:
```suggestion
                image._width, image._height = pil_image.size
```

I am not sure if `self` is correct here...

filer/management/commands/filer_check.py
```diff
@@ -112,3 +121,30 @@ def walk(prefix):
...
+                self._transparent = easy_thumbnails.utils.is_transparent(pil_image)
```
fsbraun, line 148:
```suggestion
                image._transparent = easy_thumbnails.utils.is_transparent(pil_image)
```

Same here..

fsbraun:
@benzkji I would be fine if you caught the `PIL.UnidentifiedImageError`, and potentially any `FileNotFound` error in the management command. They indicate a inconsistent state: corrupted or missing files.

benzkji:
probably last thing: test with SVG, and "corrupt" SVG.

fsbraun:
@benzkji I think, also a SVG-based test would be great. Here's a way to easily test SVGs: 

https://github.com/benzkji/django-filer/blob/6aa899527585d6d352bb0c008cf451791e168a4c/tests/test_validation.py#L26-L30

fsbraun (approved):
Great work, @benzkji ! Thanks so much!

PR #1464, opened 2024-05-17
method clean_subject_location not implemented correctly

jrief:
## Description

Method `clean_subject_location` not implemented correctly.

According to the [Django docs](https://docs.djangoproject.com/en/5.0/ref/forms/validation/#cleaning-a-specific-field-attribute) and to my own experience, a method for cleaning a specific field attribute should not itself call `super().clean()`. This is because this method is invoked later anyway. In the current implementation the method `FileAdminChangeFrom.clean()` is called twice, which in my opinion should not happen.

It also sometimes causes problems if users try to replace an image using the Filer's admin image detail view.

Since no functionality changed, I did not add any extra test.

## Checklist

### DIFF
--- a/filer/admin/imageadmin.py
+++ b/filer/admin/imageadmin.py
@@ -46,8 +46,7 @@ class ImageAdminForm(FileAdminChangeFrom):
         for subject location widget to receive valid coordinates on field
         validation errors.
         """
-        cleaned_data = super().clean()
-        subject_location = cleaned_data['subject_location']
+        subject_location = self.cleaned_data['subject_location']
         if not subject_location:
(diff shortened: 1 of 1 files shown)


fsbraun (approved):
Well spotted!

jrief:
sorry, unfortunately `clean_data['owner']` might not be set. Just found out while testing locally. `owner` isn't used anyway in `validate_upload(…)` so using `cleaned_data.get("owner")` is safer.

fsbraun:
Where's `cleaned_data["owner"]` used and how does that affect the PR? Maybe that's another PR?

PR #1467, opened 2024-05-17
fix: Folder select widget did not render correctly with standard Django admin styles

### LINKED ISSUE #1463: PROBLEM IN TEMPLATE (admin_foler.html) WHEN USING 'FilerFolderField'
Hello,
I am working on updating a Django project, which was using version 3.2.6 and consequently used django-filer 2.0.2. django-filer was used so that the user could choose a folder with information relevant to a particular field!!! In the models it was defined like this : 

imagem_1

And everything worked as expected!

imagem_2

In this project update, I updated Django to version 5.0.6 and consequently the django-filer to the latest version, 3.1.1.

When using the same field, I noticed that there are problems with the template, as shown in the image below!!!

imagem_3

I was then able to detect that the problem came from the admin_folder.html file, because when I replaced it with the html from the old version, it already allowed me to see the add folder field...
The current version of the template has the classes hidden, and I can't understand why!

I don't know if I'm missing a setting or something, but I think it's a problem with the template!!!

Your insights into resolving this performance issue would be greatly appreciated.

fsbraun:
## Description

This PR fixes #1463 .

## Related resources

* #1463 
* #...

## Checklist


fsbraun:
@jrief I hoped you'd include this in today's release...

jrief (approved):

jrief:
sorry, that was too late.

Anyway, I'll have to release 3.1.3 then today.

PR #1490, opened 2024-08-20
fix: Reduce number of thumbnails created for admin, avoid admin thumbnails for svg files

### LINKED ISSUE #1489: Uploading bigger SVG files blocks the UI
Users think the system is broken. I've had this issue several times, with architecture projects.

I wonder if the old "create 4 thumbnails for every image right during/after upload" system is still in place? This could be replaced by just creating ONE thumbnail, during rendering of the change list? Thus taking of quite a bit of load, during upload, and making everything a bit more reactive?

fsbraun:
## Description

Over the years, several thumbnail file sizes for the filer admin have accumulated:
* 16x16, 32x32, 64x64
* 40x40/80x80 (current table listing)
* 75x75 for detail views
* 160x160 (current icon listing)
* 180x180 formerly used in the admin widget

This PR reduces the thumbnail sizes to two 
* `FILER_TABLE_ICON_SIZE = 40` for table listings plus twice the value for detail views
* `FILER_THUMBNAIL_ICON_SIZE = 120` for the thumbnail view

No thumbnails are generated for SVG files shown in the admin. 

This PR fixes #1489 and reduces the amount of storage used up for thumbnails used by the filer admin interface.

Currently, the icon sizes are also hard-coded into the css. The next step will be to make them html attributes and have dynamic css that adjusts to the attributes.

## Related resources

* #1489 
* #...

## Checklist


filer/admin/folderadmin.py
```diff
@@ -271,11 +273,13 @@ def directory_listing(self, request, folder_id=None, viewtype=None):
 
         list_type = get_directory_listing_type(request) or settings.FILER_FOLDER_ADMIN_DEFAULT_LIST_TYPE
         if list_type == TABLE_LIST_TYPE:
-            size = "40x40"  # Prefetch thumbnails for listing
-            size_x2 = "80x80"
+            s = FILER_TABLE_ICON_SIZE  # Prefetch thumbnails for table view
```
vinitkumar, line 276: Can we please more descriptive variable name than just `s` here?

vinitkumar (approved):
Just one minor feedback. Otherwise, this looks excellent. Great work as always @fsbraun :)

filer/admin/folderadmin.py
fsbraun: Yeah, I fully removed it because the descriptive names are actually the settings like `FILE_TABLE_ICON_SIZE`.

vinitkumar (approved):

Issue #1495, opened 2024-09-13
Projects with custom AUTH_USER_MODEL are broken in version 3.2


payamnj:
For type hinting the user parameter in the filer/cache.py functions, a new change added to get the user model.
https://github.com/django-cms/django-filer/blob/44255c8a15f8973f34420aa0c9c869e281d9d45e/filer/cache.py#L7

Since this code can be executed before the django app is fully initialized (for example if we import FilerFileField in our application model)
it will not be able to load the custome auth user model.

`django.core.exceptions.ImproperlyConfigured: AUTH_USER_MODEL refers to model 'my_app.User' that has not been installed`

Maybe we can just use TypeVar here for type hinting the user parameters instead!

fsbraun:
@payamnj Thanks, for reporting this! Good catch.

payamnj:
@fsbraun Yes, I will create a PR for this.

PR #1496, opened 2024-09-13
fix: Use TypeVar for type-hinting of the user parameter

payamnj:
## Description

Using the AbstractBaseUser for type-hinting the auth user parameter to fix the problem mentioned in issue #1495

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/filer/cache.py
+++ b/filer/cache.py
@@ -1,13 +1,13 @@
 import typing
 
-from django.contrib.auth import get_user_model
+from django.db.models import Model
 from django.core.cache import cache
 
 
-User = get_user_model()
+UserModel = typing.TypeVar('UserModel', bound=Model)
 
 
-def get_folder_perm_cache_key(user: User, permission: str) -> str:
+def get_folder_perm_cache_key(user: UserModel, permission: str) -> str:
     """
     Generates a unique cache key for a given user and permission.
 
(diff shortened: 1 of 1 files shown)


fsbraun:
Thank you very much for this PR!

Quick question for my understanding: Why did you not use `TypeVar` as you suggested in #1495 ?

payamnj:
I wanted to know if we can be more specific, and [Django recommendation](https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#specifying-a-custom-user-model) for custom auth user model is to inherit from AbstractBaseUser, AbstractBaseUser so I thought it should work for the majority of the cases.

But since it's just a recommendation, maybe I should use the TypeVar and allow more flexibility? What do you think?

fsbraun:
Maybe with a `TypeVar` we have the most flexibility?

payamnj:
@fsbraun How about using `models.Model` instead? Any custom `AUTH_USER_MODEL` must be a subclass of Django Model, so this will provide more accurate type hinting and improve editor support.

fsbraun:
@payamnj That sounds good to me! Maybe with a comment that a user model is expected but not imported...!

fsbraun (approved):
Thanks! Great work!

payamnj:
@fsbraun can we merge this change and publish a new version?

fsbraun:
@payamnj Version 3.2.3 is on its way! Have you joined the django CMS discord community yet? https://discord-main-channel.django-cms.org/ There's a #django-filer channel, too.

payamnj:
@fsbraun No I have not yet, I will join. Thank you 🙏

PR #1519, opened 2025-03-13
fix: enhance thumbnail handling for public and private files

earthcomfy:
## Description

We use a custom naming pattern to be able to do a reverse lookup to the filename to serve protected files in `serve_protected_thumbnail`. I checked easy_thumbnail's default method of naming, and here is the difference:

Easy thumbnail: "cat.jpg.100x100_q80_crop_upscale.jpg"
Filer: "cat.jpg__100x100_q80_crop.upscale.jpg"

This PR aims to use this custom naming only for private files. Otherwise, let the easy_thumbnail package handle it. Let me know if there is anything that can be improved.

## Related resources

https://github.com/django-cms/django-filer/issues/1093

## Checklist

## Summary by Sourcery

Use standard easy-thumbnails naming for public files while retaining reversible naming and compatibility for private thumbnails.

New Features:
- Support reversible thumbnail filename resolution for both current easy-thumbnails names and legacy Filer names when serving protected files.
- Honor custom thumbnail namers for public files while ensuring private thumbnails remain resolvable and SVG thumbnails retain their source extension.

Bug Fixes:
- Preserve access to legacy private thumbnails after upgrading and avoid relying on storage lookups when determining existing admin thumbnails.

Enhancements:
- Improve directory listings by prefetching current thumbnail names and selecting standard and high-resolution icons through the configured thumbnailer.

Tests:
- Add coverage for public and private thumbnail naming, custom namers, SVG handling, legacy protected thumbnails, and filename parsing.


tests/test_models.py
```diff
@@ -35,13 +35,14 @@ def tearDown(self):
         for f in File.objects.all():
             f.delete()
 
-    def create_filer_image(self, owner=None):
+    def create_filer_image(self, owner=None, is_public=True):
```
fsbraun, line 38: Why do we need to change the signature here? Is this for testing?

filer/admin/folderadmin.py
```diff
@@ -375,8 +375,30 @@ def directory_listing(self, request, folder_id=None, viewtype=None):
...
+                thumbnail_qs.filter(name__contains=f"__{size}_").values_list("name")[:1]
+            ),
+            thumbnailx2_name_filer=Subquery(
+                thumbnail_qs.filter(name__contains=f"__{size_x2}_").values_list("name")[
+                    :1
+                ]
+            ),
+            # For public files (easy_thumbnails pattern)
+            thumbnail_name_easy=Subquery(
+                thumbnail_qs.filter(name__contains=f".{size}_q85_crop").values_list(
```
fsbraun, line 389: Indeed, `_q85`  or `crop` imho should not be part of the query. 

Can this all be simplified by customizing the part before `{size}` (either `.` or `__`). Also, how would reversability be affected if we also did use the `__` characters before size? Then this would dramatically simplify.

fsbraun (commented):
Hi @earthcomfy ! Great start!

How about changing the filer implementation for private files to match the one for public files? This would lead to less cases needed to be looked at. The key difference is that the public ones would be configurable through easy-thumbnails settings the private onces not to keep it reversable.

Probably the regex would need to be adjusted, too.

What do you think?

earthcomfy:
@fsbraun I have updated. Let me know what you think

filer/utils/filer_easy_thumbnails.py
```diff
@@ -19,58 +18,41 @@ def thumbnail_to_original_filename(thumbnail_name):
...
+            is_public = not isinstance(self.thumbnail_storage, PrivateFileSystemStorage)
 
-        basedir = self.thumbnail_basedir
-        subdir = self.thumbnail_subdir
-
-        # make sure our magic delimiter is not used in all_opts
-        all_opts = all_opts.replace('__', '_')
-        filename = f'{source_filename}__{all_opts}.{extension}'
+        path, source_filename = os.path.split(self.name)
+        thumbnail_name = super(ThumbnailerNameMixin, self).get_thumbnail_name(
```
fsbraun, line 37: I fear this call will look up the thumbnail namer setting and call that function. I see two ways to avoid that:
* Patch `self` to use the default thumbnail namer (e.g., using a context manager) if the file is not public
* Copy the current logic into filer

fsbraun:
@earthcomfy This already looks much cleaner, but I see an issue here: https://github.com/django-cms/django-filer/pull/1519/files#r2005356189

filer/utils/filer_easy_thumbnails.py
earthcomfy: @fsbraun My idea behind this was that we get all their optimizations and future improvements of easy thumbnails. This way we modify the returned value for private files (as verified in the tests).

fsbraun:
@earthcomfy One thing that caught my attention is easy-thumbnail's habit of lowercasing all file names. Does that impact reversibility for the secure media files?

Also, I see one potential issue for secure media files:
* Can we guarantee only thumbnail names are being reversed?
* If I upload a thumbnailed file as an original to secure media, the original technically can be reversed to a non-existing original.

Intuitively I believe this is ok, but I wanted to check with you.

Finally, the `filer_check` command does not remove old thumbnails. I think we have to add some information to the readme that the change will lead to regeneration of all thumbnail files and that we recommend to either delete all old thumbnails or -if that is no option - can we give a thumbnail namer that corresponds to the old filer namer.

fsbraun:
Thanks for this, and sorry it sat for so long. The goal is right and I want it in: public files should honour `THUMBNAIL_NAMER`, private files have to keep a name that `serve_protected_thumbnail()` can turn back into the source path. I did a review pass and found a few things that need fixing before it can be merged, and I've prepared the fixes locally on top of a merge with current `master`.

## What needed fixing

**1. SVG thumbnails were written as `.jpg`/`.png` but contained SVG markup.**

The old `ThumbnailerNameMixin` forced `extension = source_extension` when the source was an SVG. `Thumbnailer.get_thumbnail_name()` has no such rule — it only honours `THUMBNAIL_PRESERVE_EXTENSIONS`, which filer never sets — while `generate_thumbnail()` still branches on the *source* name and calls `engine.save_svg_image()`. On the branch:

```
master: …/probe.svg__100x100.svg
PR:     …/probe.svg.100x100_q85.jpg   ← contains b'<svg width="10.0" …'
```

So every SVG thumbnail (admin detail view, `Image.thumbnails`, `icon_view`) got served as `image/jpeg` with XML inside. The `transparent=True` path made it `.png` with the same problem.

**2. Private-vs-public was detected by sniffing the storage class name.**

```python
is_public = "PrivateFileSystemStorage" not in str(self.thumbnail_storage.__class__)
```

`FILER_STORAGES['private']['thumbnails']['ENGINE']` is a documented setting that people override (django-storages S3, a custom subclass). With any such backend a private file is classified as public, gets `THUMBNAIL_NAMER`, and then `thumbnail_to_original_filename()` can't recover the source path — so every private thumbnail 404s. That's exactly the invariant the PR is trying to protect.

**3. No backwards compatibility for thumbnails already on disk.**

The new regex strips the last two dot-segments, so an old-scheme name resolves wrongly:

```python
thumbnail_to_original_filename("ab/cd/foo.jpg__200x200_q85.jpg")  # → "ab/cd/foo"
```

`File.objects.get(file=…)` then raises `DoesNotExist` and `serve_protected_thumbnail()` returns 404 for every already-generated private thumbnail, including URLs that are already out there.

**4. The directory-listing annotation defeated the PR's own feature.**

`name__contains=f".{size}_"` only matches names from `easy_thumbnails.namers.default`. With `namers.hashed` (`6qW1buHgLaZ9.jpg`), `source_hashed`, or a custom namer, `thumbnail_name` is always `NULL`, `file_icon_context()`'s `configured_name == file.thumbnail_name` check always fails, and every image in every listing falls through to a `admin:filer_file_fileicon` redirect — one extra admin request per image, and no `highres_url` at all.

**5. Two of the new tests didn't discriminate.**

`assertRegex(name, r"^.*\..*\.[^.]+$")` is satisfied by almost any path with two dots, including the custom namer's output, so `test_thumbnailer_class_for_public_files` / `..._private_files` would pass with the namer selection inverted. Also `custom_namer` returns `os.path.join(path, …)` — a full path — which easy-thumbnails joins under `basedir/path/subdir` again, duplicating the directory; the test only checked `os.path.basename`, so it hid that.

## What I changed

* **SVG:** the naming override now also sets `thumbnail_preserve_extensions = True` when the source is an SVG, so the name keeps `.svg` and matches what `generate_thumbnail()` actually writes.
* **Private detection:** `is_public` is passed in explicitly from `BaseImage.easy_thumbnails_thumbnailer`, with `self.instance.is_public` (for `MultiStorageFieldFile`) and the storage's `is_secure` attribute as fallbacks. No more class-name matching.
* **Legacy names:** `original_filename_candidates()` recognises both `source.jpg__100x100_q85.jpg` and `source.jpg.100x100_q85.jpg`. Both schemes start the options segment with `<width>x<height>`, which disambiguates a `__` that is part of the source filename. `serve_protected_thumbnail()` tries each candidate, so existing private thumbnail URLs keep working.
* **Directory listing:** replaced the two correlated subqueries with a single prefetch of the current page's thumbnail names (`FolderAdmin._attach_thumbnail_names`), compared in full by the template tag. That works with any namer, only touches the files on the page instead of the whole folder, and lets the `exclude(name__contains="upscale")` TODO go. Net one extra query, so `test_filer_directory_listing_performance` goes 7 → 8.
* **Tests:** exact names instead of loose regexes, `custom_namer` fixed, and new coverage for SVG generation, namer selection, restoring the namer options, and the legacy URLs. `tests/test_filer_easy_thumbnails.py` now uses a real `FilerThumbnailer` — the mixin can no longer be used standalone now that it delegates to `Thumbnailer`.
* **CHANGELOG:** entry plus an upgrade warning, since thumbnail file names change and the old files stay behind in the thumbnail storage.

The branch was also ~250 commits behind, so I merged `master` in first; `tests/test_filer_easy_thumbnails.py` and `tests/test_utils.py` on `master` cover this module directly and needed updating alongside the change. Full suite (460 tests) and ruff are green.

Happy to push this to your branch if you enable maintainer edits, or I can open a follow-up PR against it — whichever you prefer.

One thing I deliberately left out: there's no cleanup path for the public thumbnails orphaned by the rename. They're just regenerated under the new name and the old files stay on disk. Worth a separate issue.

PR #1538, opened 2025-08-25
fix: Folder permission cache update sometimes raised TypeError

### LINKED ISSUE #1537: folder permission cache update sometimes raises `TypeError: 'set' object does not support item assignment`
With this test case the error can be reproduced:

```python
    def test_update_folder_permission_cache_overwrites_existing_cache_value(self):
        update_folder_permission_cache(self.user, self.permission, {2})
        update_folder_permission_cache(self.user, self.permission, self.id_list)  # Error here
        permissions = get_folder_permission_cache(self.user, self.permission)
        self.assertEqual(permissions, self.id_list)
```

I noticed this, as it is happening on out production servers.

We are currently using version `3.2.3` of `django-filer` but it is still an issue in the latest version on the `master` branch.

fsbraun:
## Description

This PR fixes #1537. Thanks to @PeterW-LWL for catching this and providing a regression test!

@PeterW-LWL Do you want to give it a test run?

## Related resources

* #...
* #...

## Checklist

## Summary by Sourcery

Fix folder permission cache updating logic to store permissions as a set and overwrite stale cache values, preventing the observed TypeError.

Bug Fixes:
- Fix update_folder_permission_cache to always overwrite the cache with a set to prevent TypeError

Tests:
- Add regression test to verify that update_folder_permission_cache overwrites existing cache entries

### DIFF
--- a/filer/cache.py
+++ b/filer/cache.py
@@ -86,4 +86,4 @@ def update_folder_permission_cache(user: UserModel, permission: str, id_list: ty
     permission (str): The permission to update.
     id_list (list): The list of IDs to set as the new permissions.
     """
-    cache.set(get_folder_perm_cache_key(user, permission), {user.pk: set(id_list)})
+    cache.set(get_folder_perm_cache_key(user, permission), {user.pk: id_list})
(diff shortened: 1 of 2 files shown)


PeterW-LWL:
Hey @fsbraun, thanks for the fast reaction.

This pull request looks good. I'll try this out.

But I noticed that some of the docs and typings are not correct in the `filer/cache.py` and `tests/test_permission_cache.py:14`. Every where the code is used with `set` and not with `list`. Also `get_folder_permission_cache` does not return a `dict`.

PR #1571, opened 2026-01-15
fix: infinite recursion bug on django-polymorphic >=4.9

### LINKED ISSUE #1569: bug: Infinite recursion with django-polymorphic>=4.9
When using django-polymorphic>=4.9 the tests start failing with infinite recursion:

```
...
  File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/django/db/models/sql/where.py", line 218, in clone
    clone = self.create(connector=self.connector, negated=self.negated)
  File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/django/utils/tree.py", line 36, in create
    obj = Node(children, connector or cls.default, negated)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

bckohan:
## Description

closes #1569

There are a few things going on here:

1. Prior to django-polymorphic 4.9 the ``base_manager`` (used for loaddata and related fields) was set to the same as the ``default_manager`` by default. This was incongruous with Django's behavior so it was fixed in [django-polymorphic/#816](https://github.com/jazzband/django-polymorphic/pull/816).
2. A custom ``PolymorphicQuerySet`` is used on ``File`` objects as the default manager. It overrides ``only()`` to always add a few fields including ``_file_size`` and ``is_public`` because these fields are accessed in the model's ``__init__`` method. I suspect this infinite recursion bug was encountered before, hence the override.
3. If you delete a ``Folder`` object, internally Django does something like: ``folder.all_files.only("id").delete()``. This triggered the bug because it was now using a plain ``PolymorphicManager`` that did not add the extra fields to ``only()``.
4. The fix, implemented in this PR is to set ``base_manager_name = "objects"`` to force related managers to use the overridden ``only()`` implementation.
5. There was *another* bug in django-polymorphic 4.9, 4.10.0-4.10.2 that prevented Meta.base_manager_name from being respected. This has been fixed in 4.9.1 and 4.10.3. The offending releases were yanked.

This PR, combined with the fixes in 4.9.1 and 4.10.3 should fix everything. It will also work on versions prior to 4.9.

**You probably want to merge this PR soon because if anyone installs django-polymorphic 4.9+ with the existing release they will not be able to delete Folders that aren't empty.** I am also a user of this great library so thank you!!

I also restricted the django-polymorphic dependency to < 5. As the maintainer I can attest that there will be breaking changes (but for great reasons!)

## Related resources

* This is the relevant PR: https://github.com/jazzband/django-polymorphic/pull/816 on django-polymorphic.
* I think this issue is also related: https://github.com/jazzband/django-polymorphic/issues/334
* https://github.com/jazzband/django-polymorphic/issues/844
* Minimal reproduction: https://github.com/jazzband/django-polymorphic/issues/845

## Checklist

## Summary by Sourcery

Update filer File model manager configuration to align with newer django-polymorphic behavior and prevent recursion issues when deleting related objects.

Bug Fixes:
- Ensure related managers for File use the custom default manager by setting base_manager_name to the objects manager, preventing infinite recursion when deleting folders with newer django-polymorphic versions.

Build:
- Relax django-polymorphic dependency to allow all 4.x releases while remaining below 5.0.

Chores:
- Add a Django migration to persist the updated File model options in the database schema history.


bckohan:
This PR illustrates that it was the base manager change that triggers the bug. I've marked this as a WIP because I haven't figured out why the bug occurs when you use a basic PolymorphicManager. Its something involving the is_public flag.

PR #1630, opened 2026-09-03
feat: Add default storages, and create public ambit in a migration

fsbraun:
## Summary by Sourcery

Provide usable default Finder storages and an initial ambit while securing and simplifying browser API access.

New Features:
- Derive public and public-sample Finder storages from a filesystem default when they are not explicitly configured.
- Create a default public ambit during migration, with configurable slug and opt-out behavior.
- Serve Finder browser API endpoints through the admin while supporting project-level URL mounts.

Bug Fixes:
- Enforce folder and inode permissions consistently across browser API operations, including anonymous access and indistinguishable handling of missing resources.
- Recover safely from stale or cross-ambit browser session folder references.
- Prevent Finder admin customization from affecting unrelated application indexes.

Enhancements:
- Add configuration system checks for invalid default ambit slugs and unavailable storage aliases.
- Improve ambit management command storage defaults and storage descriptions.
- Expand user-facing documentation for storage configuration, default ambits, and ambit management.

Documentation:
- Document automatic public storage derivation, default ambit creation, configuration settings, and manual ambit setup.

Tests:
- Add coverage for default ambit migration behavior, derived storages, admin API routing, browser authorization, and stale browser state handling.

Chores:
- Restrict anonymous users from matching everyone ACL entries to prevent unauthenticated browser access.

## Related resources

* #...
* #...

## Checklist

* [ ] I have read the [contribution guidelines ](https://github.com/django-cms/django-filer/blob/master/docs/development.rst#contributing) and I have joined #pr-reviews on
[Discord](https://www.django-cms.org/dicord) to find a “pr review buddy” who is going to review my pull request.


finder/browser/views.py
```diff
@@ -110,12 +158,30 @@ def structure(self, request, slug=None):
...
+
+        Refused the same way as an unknown ambit, so that neither the existence of an
+        inode nor the reason for the refusal can be read off the response.
+        """
+        try:
+            inode = FileModel.objects.get_inode(id=inode_id, **lookup)
+        except ObjectDoesNotExist:
+            logger.warning("No inode “%s”.", inode_id)
+            raise PermissionDenied(self.access_denied_message)
+        folder = inode if inode.is_folder else inode.folder
```
jrief: This is redundant. Models `File` and `Folder` already offer a property method named `folder` which does exactely that.

finder/migrations/0002_default_ambit.py
jrief, line 1: In order to start clean with version 4, I actually would prefer to merge this into `0001_initial.py`.

finder/models/folder.py
```diff
@@ -184,6 +186,10 @@ def as_dict(self, ambit):
         }
 
     def has_permission(self, user, privilege):
```
jrief, line 188: I actually would consider any Folder/File/Inode without a realy user, a runtime error. This must never happen. I therefore would add an assert statement here, or do you have strong opinions against this?

On the other side, the FK `owner`may be `None`. Is this a unintended as well?

finder/models/inode.py
```diff
@@ -361,6 +361,8 @@ def get_meta_data(self):
 
     def has_permission(self, user, privilege):
         assert (privilege & Privilege.FULL) != 0, "Invalid privilege value."
```
jrief, line 363: same here

jrief (changes requested):
let's talk about these chane requests personally.

finder/browser/views.py
```diff
@@ -158,12 +158,30 @@ def structure(self, request, slug=None):
             **self.list(request, last_folder_id),
         }
 
+    def _get_inode(self, request, inode_id, privilege, **lookup):
```
jrief: Since this is a private utility method, I would move it directly below `_get_can_view_subquery`

finder/browser/views.py
fsbraun: Good catch, fixed. It reads `inode.folder.has_permission(...)` now.

finder/migrations/0002_default_ambit.py
fsbraun: Done — 0002 is gone and the `RunPython` is the last operation of `0001_initial`. The module docstring became the docstring of `create_default_ambit`.

finder/browser/views.py
fsbraun: Moved. One thing to flag: it now sits just above the `access_denied_message` class attribute that both it and `_get_ambit` use.

finder/models/folder.py
fsbraun: `is_anonymous(user)` is `user is None or user.is_anonymous`, and the two halves are not the same kind of thing:

* `user.is_anonymous` is load-bearing at runtime. `get_urls` in `finder/admin/ambit.py` mounts the browser endpoints without `admin_view()` on purpose — the widgets render on ordinary forms outside the admin, so a non-staff user has to be able to reach them. An unauthenticated request therefore arrives at `BrowserView` by design, with `request.user` an `AnonymousUser`. In `InodeModel.has_permission` the "everyone" ACE is `Q(user__isnull=True, group__isnull=True)`, which an `AnonymousUser` matches — **so that guard is the only thing between an anonymous visitor and READ_WRITE on the default ambit the new migration creates**. As an assert it disappears under `python -O,` which turns a security check into a no-op in exactly the deployment where it matters most.
* `user is None` is the half that really should never happen. Every call site passes `request.user`, and with `AuthenticationMiddleware` installed that is `AnonymousUser`, never None.

So I'd keep the runtime check. If you want the impossible case to be loud, I'd rather narrow is_anonymous to user.is_anonymous and assert on user is not None separately, than fold both into one assert.

On your second question — owner being nullable seems intentional to me. It's `null=True`, `on_delete=SET_NULL`, so deleting a user doesn't cascade-delete their files and orphan the payloads in storage. Making it non-nullable would mean picking a different `on_delete`, and every option there looks worse to me.

finder/models/inode.py
fsbraun: This one is the actual authorization boundary: without it an `AnonymousUser` matches the "everyone" ACE (`user__isnull=True`, `group__isnull=True`) and gets whatever that entry grants — READ_WRITE, for the ambit the new migration creates. `FolderModel`'s override guards the `is_trash` and `self.owner == user` shortcut above `super()`.

jrief (approved):

